PR #12, opened 2025-09-08
Feature/zero copy pathing

drondeseries:


drondeseries:
Download client altmount places downloads in the root folder /mnt/usenet-rclone/tv. You should not download to a root folder.

Not a good practice

PR #52, opened 2025-10-05
Update manager.go

drondeseries:
Explanation of the Fix
Original Code:

Go

func (c *Config) GetActualMountPath(provider string) string {
    return filepath.Join(c.MountPath, provider)
}
Corrected Code:

Go

func (c *Config) GetActualMountPath(provider string) string {
    // FIX: Do not join the provider name. Return the MountPath directly.
    return c.MountPath
}
The original function was taking your mount_path and appending the hardcoded provider name "altmount". The corrected function now simply returns the mount_path directly, which is the behavior you want.

### DIFF
--- a/internal/config/manager.go
+++ b/internal/config/manager.go
@@ -717,9 +717,10 @@ func (c *Config) ToNNTPProviders() []nntppool.UsenetProviderConfig {
 	return providers
 }
 
-// GetActualMountPath returns the actual mount path used by rclone, which includes the provider subdirectory
+// GetActualMountPath returns the actual mount path used by rclone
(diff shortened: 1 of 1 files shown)


javi11:
If you do that, the mounting poing will be for instance /mnt/remote, instead to /mnt/remote/altmount, and rclone is mounting in /mnt/remote/altmount, would´t that be the right path? or you want to get ride of the hardcoded altmount?

drondeseries:
Getting rid of that hard coded altmount, it's conflicting with the arrs and sabnzbd path

drondeseries:
Conflict with internal rclone

javi11:
thanks for the PR I will do the changes :)

PR #65, opened 2025-10-19
for review

MythodeaLoL:
fix for error loop? 

fix for `http: superfluous response.WriteHeader call from /internal/webdav.NewHandler.func2 (adapter.go:137)`


MythodeaLoL:
error sending

PR #98, opened 2025-11-17
Fix: Remove recursive parameter from vfs/forget call and enhance vfs/…

drondeseries:
…refresh logic


javi11:
Hey @drondeseries what are you try to solve by calling forget first and then refres?

PR #103, opened 2025-11-26
fix(par2): prevent panic and queue stuck from corrupted PAR2 files

sureshfizzy:
Fix "makeslice: len out of range" panic that caused queue items to get stuck and the application to crash when processing corrupted PAR2 files with invalid packet lengths.

Add validation checks before slice allocation to prevent crashes:
 - Validate packet sizes (max 1GB) and filename lengths (max 64KB)
 - Return errors for corrupted files instead of panicking

This prevents the application from crashing and allows queue processing to continue normally.


javi11 (approved):

javi11:
@sureshfizzy can you fix the error on internal/usenet/validation.go:8:2: "time" imported and not used

PR #109, opened 2025-12-04
fix(par2): prevent panic and queue stuck from corrupted PAR2 files

sureshfizzy:
Fix "makeslice: len out of range" panic that caused queue items to get stuck and the application to crash when processing corrupted PAR2 files with invalid packet lengths.

Add validation checks before slice allocation to prevent crashes:
 - Validate packet sizes (max 1GB) and filename lengths (max 64KB)
 - Return errors for corrupted files instead of panicking

This prevents the application from crashing and allows queue processing to continue normally.

### DIFF
--- a/internal/api/health_handlers.go
+++ b/internal/api/health_handlers.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/gofiber/fiber/v2"
+	"github.com/javi11/altmount/internal/config"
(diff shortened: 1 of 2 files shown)


sureshfizzy:
@javi11 I think we are good now, but I see some other issue on workflow

MythodeaLoL:
main branch have some build/test error too, on ARM build.

PR #111, opened 2025-12-05
Alpha5

drondeseries:


MythodeaLoL:
Hi @drondeseries, could you talk a little about the proposed changes, as is done in other pull requests? This way we can better understand the key points that have been improved, removed, and new implementations.

altmount
MythodeaLoL: You can leave this file out of the pull request.

MythodeaLoL (changes requested):

Issue #115, opened 2025-12-06
New Setup - Authentication issue


rottstein:
Just installed and got setup with admin account but every page has "Something went wrong   Authentication required"

No clue how to move forward here, cant configure or anything.

![Image](https://github.com/user-attachments/assets/58091c14-95a7-4fb7-85de-14be132cbfc2)

MythodeaLoL:
Follow https://javi11.github.io/altmount/

rottstein:
Did that.  Think I got as far as I did by just winging it?

kingciko:
İ have the same problem. İ did that what's writen in the documentation. Last week I try it and it work and today I want to make the settings on my server it don't work

MythodeaLoL:
here too, incredibly everything was normal until a few minutes ago, I recreated the configuration and the same problem occurred as with you. 😮

old builds have same problem, we now know that it's something related to the previous JWT authentication codes.
I informed Javi about this.

MythodeaLoL:
Fixed.

PR #116, opened 2025-12-06
fix(auth): Add JWT cookie configuration options

MythodeaLoL:
try to fix, untested #115


MythodeaLoL:
problem continue
image

PR #120, opened 2025-12-09
Fix docs for Windows rclone commands

hansmok:
Updated the docs to reflect some of the things I needed to do in order to get this working on a Windows setup; I assume that the paths and commands were probably changed and but not reflected in the docs.

### DIFF
--- a/docs/docs/1. intro.md
+++ b/docs/docs/1. intro.md
@@ -166,19 +166,19 @@ AltMount includes a built-in rclone mount feature for the WebDAV interface:
 >
 > # Then create the config:
 > rclone config create altmount webdav `
->   url http://localhost:8080 `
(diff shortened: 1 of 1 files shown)


MythodeaLoL:
and what about the documentation for Linux? because Alt is designed more for Linux than for Windows.

hansmok:
AltMount in the container feels pretty platform agnostic to me; I don't know that I agree with the sentiment that it's "designed" for a specific platform. But to answer the question I don't have a Linux host for personal use so I wouldn't know what to update since I haven't gone through the process; the only thing that I ran into would be /mnt/altmount not being created properly since the image didn't run with high enough level perms to create the file in the root.

javi11:
This will works for linux too

PR #122, opened 2025-12-09
feat(migration): add a tool to migrate nzbdav to altmount

drondeseries:
feat(migration): add a tool to migrate nzbdav to altmount
fix(path): fix meta creation for one file only, now a folder is not created if the name matches
fix(import): fail if there is no files processed
feat(import): allow extension with . to be specified in allowed filter

### DIFF
--- a/internal/nzbdav/parser.go
+++ b/internal/nzbdav/parser.go
@@ -115,33 +115,33 @@ func (p *Parser) deriveRelPath(path, category string) string {
 	path = strings.Trim(path, "/")
 
 	// 2. Identify and remove category prefix
-	// Simple heuristic: check if path starts with known category roots
-	// We use the derived category to guess the prefix
+	// We want to find the category folder in the path and return everything after it
+	// e.g. /content/tv/Show/Season 1 -> Show/Season 1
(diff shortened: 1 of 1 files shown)


internal/metadata/service.go
```diff
@@ -60,9 +60,25 @@ func (ms *MetadataService) WriteFileMetadata(virtualPath string, metadata *metap
 		return fmt.Errorf("failed to marshal metadata: %w", err)
 	}
 
-	// Write to file
-	if err := os.WriteFile(metadataPath, data, 0644); err != nil {
-		return fmt.Errorf("failed to write metadata file: %w", err)
+	// Create a temporary file in the same directory to ensure atomic rename
+	tempFile, err := os.CreateTemp(metadataDir, truncatedFilename+".*.tmp")
```
javi11: Why this part?

drondeseries: this code ensures that metadata files are
  written reliably. If the system crashes or the write operation is   interrupted, the metadata file won't be corrupted or left in an     incomplete state. This is especially important for altmount to      correctly track files, their health, and their associated NZBs.

PR #126, opened 2025-12-11
feat: add speed test

drondeseries:
- **feat: refactor nzbdav import to be asynchronous with progress tracking**
- **feat: add top speed metrics and test file download with max connections**


internal/importer/service.go
```diff
@@ -623,7 +816,17 @@ func (s *Service) processNzbItem(ctx context.Context, item *database.ImportQueue
 		basePath = filepath.Join(basePath, *item.Category)
 	}
 
-	return s.processor.ProcessNzbFile(ctx, item.NzbPath, basePath, int(item.ID))
+	// Determine if max connections override is needed
+	var maxConnectionsOverride *int
```
javi11: Why do you want to override the connections here? This can be dangerous and a potentialy cpu and memory leak

PR #132, opened 2025-12-14
feat(fuse): Add native FUSE mount managed by UI

drondeseries:
This PR introduces a native FUSE mount for AltMount, fully managed via the Web UI. It includes:
- FUSE service implementation (go-fuse/v2)
- Graceful restart logic
- Configurable UID/GID
- Correct file mode handling
- UI for configuration
- Removal of CLI command

Resolves issues related to direct filesystem access and allows seamless integration for media servers.

### DIFF
--- a/internal/fusefs/file.go
+++ b/internal/fusefs/file.go
@@ -78,15 +78,9 @@ func (f *NzbFile) Open(ctx context.Context, flags uint32) (fh fs.FileHandle, fus
 // Read implements fs.FileHandle.Read.
 // It reads data from the underlying afero.File.
 func (fh *NzbFileHandle) Read(ctx context.Context, dest []byte, off int64) (fuse.ReadResult, syscall.Errno) {
-	// Seek to the correct offset
-	_, err := fh.aferoFile.Seek(off, io.SeekStart)
-	if err != nil {
-		slog.Error("Failed to seek file", "offset", off, "error", err)
-		return nil, syscall.EIO
-	}
-
-	// Read data into the destination buffer
-	n, err := fh.aferoFile.Read(dest)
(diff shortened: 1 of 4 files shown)


internal/fusefs/file.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	return &NzbFileHandle{aferoFile: aferoFile}, fuse.FOPEN_KEEP_CACHE, fs.OK
+}
+
+// Read implements fs.FileHandle.Read.
+// It reads data from the underlying afero.File.
+func (fh *NzbFileHandle) Read(ctx context.Context, dest []byte, off int64) (fuse.ReadResult, syscall.Errno) {
+	// Seek to the correct offset
+	_, err := fh.aferoFile.Seek(off, io.SeekStart)
```
javi11: you don't need to seek on every read just remove this and do the read

internal/fusefs/file.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Open implements fs.Inode.Open.
+// It opens the file using MetadataRemoteFile and returns an NzbFileHandle.
+func (f *NzbFile) Open(ctx context.Context, flags uint32) (fh fs.FileHandle, fuseFlags uint32, errno syscall.Errno) {
+	// Only allow read-only access
+	if flags&syscall.O_ACCMODE != syscall.O_RDONLY {
+		return nil, 0, syscall.EACCES
+	}
+
+	ok, aferoFile, err := f.metadataRemoteFile.OpenFile(ctx, f.path)
```
javi11: The current filesystem is not prepared to use seek because it works by Range requests, if fuse needs seek you will need to modify it in metadata_remote_file

PR #135, opened 2025-12-15
fix: preserve library paths and sync metadata status during file corruption

sureshfizzy:
Fixes these bugs:
1. Library path comparison now checks both full and relative paths
2. Use COALESCE to prevent NULL from overwriting existing library paths
3. Update metadata file status when registering corrupted files

This ensures:
- Library paths (from Sonarr/Radarr) are never lost during corruption
- Healthy files automatically reappear in WebDAV mount
- Radarr/Sonarr can successfully trigger repairs with correct library paths


javi11:
Is this still needed @sureshfizzy ?

sureshfizzy:
Sorry i was at vacation XD, part of a fix might be required. I'll check with your latest changes and raise a new PR.

PR #178, opened 2025-12-29
fix(health): allow periodic re-checks for healthy files

drondeseries:
This change removes 'healthy' from the exclusion list in GetUnhealthyFiles, ensuring that files marked as healthy are picked up for their scheduled periodic re-checks. This restores the intended tiered scheduling logic.


javi11:
I think checking all items every 2 hours can be overkill for largue library, it is better release data * 2 then grow?

drondeseries:
It follows their set schedule.

PR #182, opened 2025-12-29
fix(arrs): improve category handling for ARR auto-setup and connectivity tests

drondeseries:
This PR fixes the '400 MovieCategory : Category does not exist' error during Radarr/Sonarr auto-setup.
- Added 'Category' field to ARR instance configuration API and UI.
- Updated connectivity tests to respect the configured category instead of hardcoded defaults.
- Added 'movies' and 'tv' categories to AltMount's default configuration so they exist for first-time setup.


MythodeaLoL:
Tested, and is working now. 🚀 
image

PR #217, opened 2026-01-05
fix(importer): prevent virtual paths from exposing internal .nzbs directory

drondeseries:
## Description
This PR fixes a bug where the internal persistent storage path for NZB files (e.g., `/config/.nzbs`) was incorrectly being included in the virtual directory structure for imported files.

## The Bug
When an NZB file is imported, it is moved to a persistent `.nzbs` folder. If the relative path for the import was determined to be the root (`/`) or empty (e.g., from NZBDav imports or root watch folders), `CalculateVirtualDirectory` would resolve the virtual path relative to the file's physical location.

This resulted in incorrect virtual paths leaking the internal storage structure, such as:
`metadata/config/.nzbs/Category/ShowName...`

## The Fix
Updated `processNzbItem` in `internal/importer/service.go` to:
1. Detect when a file path is within the persistent `.nzbs` folder.
2. If the calculated virtual directory exposes this internal folder name, reset the virtual directory to the intended base path (logical root or category root).

This ensures files moved to persistent storage appear correctly at the root of the library (or their assigned category) in the virtual filesystem, keeping the internal storage implementation hidden.

### DIFF
--- a/internal/importer/service.go
+++ b/internal/importer/service.go
@@ -545,6 +545,27 @@ func (s *Service) processNzbItem(ctx context.Context, item *database.ImportQueue
 	// For NZBDav: returns "/root" (e.g. "/nzb")
 	virtualDir := filesystem.CalculateVirtualDirectory(item.NzbPath, basePath)
 
+	// Fix for issue where files moved to persistent .nzbs directory end up with exposed .nzbs path in virtual directory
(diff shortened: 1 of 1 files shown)


drondeseries:
Closing in favor of a clean PR with only the relevant fix.

PR #221, opened 2026-01-05
fix(importer): use relative paths in directory watcher to prevent absolute path exposure in virtual directory

drondeseries:
## Description
This PR fixes a bug where files imported via the directory watcher would include the full absolute path of the watch directory in their virtual mount path (e.g., `/mnt/data/watch/movie.mkv` instead of `/movie.mkv`).

## The Issue
The issue was caused by the interaction between the directory watcher, the queue service, and the `CalculateVirtualDirectory` function:

1. The **Directory Watcher** calculated the `relativePath` as the absolute path to the watch root (e.g., `/mnt/data/watch`) when no category was matched, or constructed an absolute path for categories.
2. The **Importer Service** moves the NZB file from the watch directory to a persistent storage location (e.g., `/config/.nzbs/movie.nzb`) to ensure data safety.
3. `CalculateVirtualDirectory` attempts to calculate the relative path between the *new* NZB location (`/config/.nzbs/movie.nzb`) and the provided `relativePath` (`/mnt/data/watch`).
4. Since the file is no longer inside the `relativePath` (it was moved), `filepath.Rel` fails (or returns a path starting with `..`).
5. As a fallback, `CalculateVirtualDirectory` returns the `relativePath` itself, which is the full absolute path.

## The Fix
The directory watcher now passes the **relative path of the file** (relative to the watch root) as the `relativePath` to the queue, instead of the absolute watch root.

- **Before:** `relativePath` = `/mnt/data/watch` -> Virtual Path: `/mnt/data/watch`
- **After:** `relativePath` = `Movie/movie.nzb` (or similar) -> Virtual Path: `/Movie/movie.nzb`

This aligns the behavior with the API import (`/api/import/file`), which already correctly passes a relative path string that serves as the intended virtual structure.

## Verification
- Validated that `CalculateVirtualDirectory` correctly returns the relative string when `Rel` fails (simulating the move).
- Verified that category logic correctly handles the relative path input.

### DIFF
--- a/internal/importer/scanner/watcher.go
+++ b/internal/importer/scanner/watcher.go
@@ -184,7 +184,6 @@ func (w *Watcher) processNzb(ctx context.Context, watchRoot, filePath string) er
 			relativePath = &relPath
 		}
 	}
-	}
 
 	// Add to queue
 	priority := database.QueuePriorityNormal
@@ -196,7 +195,7 @@ func (w *Watcher) processNzb(ctx context.Context, watchRoot, filePath string) er
 	w.log.InfoContext(ctx, "Added watched NZB to queue",
 		"file", filePath,
 		"category", category,
-		"queue_id", item.ID)
+		"nzb_path", item.NzbPath)
 
 	// Note: We don't delete the file here because AddToQueue (Service.processNzbItem) 
 	// handles moving/renaming the NZB to persistent storage.


drondeseries:
Closing in favor of cleaner isolated branch PR #232

PR #226, opened 2026-01-06
fix(config): ensure library_sync_interval_minutes: 0 is visible in UI

drondeseries:
## Description
This PR fixes an issue where setting `library_sync_interval_minutes` (and other integer fields in HealthConfig) to `0` would cause them to be omitted from the JSON response due to the `omitempty` tag. This prevented the frontend UI from correctly displaying or handling these fields when disabled (set to 0).

## Changes
- Removed `omitempty` tag from integer fields in `HealthConfig` struct in `internal/config/manager.go`.
- Updated `config/config.yaml` to set `library_sync_interval_minutes` to 0.

## Motivation
To allow users to explicitly disable library sync (by setting it to 0) and have this configuration correctly reflected and editable in the UI.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -22,7 +22,7 @@ health:
     max_connections_for_health_checks: 5
     max_concurrent_jobs: 1
     segment_sample_percentage: 5
-    library_sync_interval_minutes: 360
+    library_sync_interval_minutes: 0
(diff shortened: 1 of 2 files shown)


drondeseries:
Closing in favor of cleaner isolated branch PR #233

PR #234, opened 2026-01-07
feat(arrs): add automatic queue cleanup for failed/stuck imports

drondeseries:
### Summary
This PR introduces a new worker to automatically clean up stuck items from Sonarr and Radarr queues. This is particularly useful for maintaining a healthy queue when using `altmount` as a download client, where some imports might fail permanently due to mapping issues or file validity checks.

### Key Changes
- **New Worker**: Added `internal/arrs/worker` that periodically checks the queue of all enabled Sonarr and Radarr instances.
- **Cleanup Logic**:
  - Removes items with "Automatic import is not possible" status (configurable via `cleanup_automatic_import_failure`, default `false`).
  - Removes items with "Episode ... was not found in the grabbed release" status (common mapping error).
  - Removes items with "No files found are eligible", "Sample file", "is not a valid video file", etc.
- **Diagnostic Logging**: Logs warnings for stuck items that match the criteria but are not eligible for auto-cleanup, aiding in debugging and future improvements.
- **Config**: Added `QueueCleanupIntervalSeconds` and `CleanupAutomaticImportFailure` to `Arrs` config.

### Motivation
Users were experiencing items stuck in "Downloaded - Waiting to Import" state that required manual intervention to remove. This feature automates that maintenance task.

PR #235, opened 2026-01-07
fix(arrs): disable blocklisting in queue cleanup

drondeseries:
Disables blocklisting when cleaning up stuck queue items to prevent false positives where successful imports (that trigger transient warnings) are incorrectly marked as failed releases.


drondeseries:
Closing in favor of updating PR #234

PR #237, opened 2026-01-08
fix(queue): remove stale item threshold check on startup

drondeseries:
Previously, the system claimed to wait 10 minutes before resetting stuck processing items, but this logic was inconsistent and could leave items stranded after a quick restart.

This change ensures that ALL items in the 'processing' state are reset to 'pending' immediately upon service startup, guaranteeing no jobs are left orphaned.


drondeseries:
Closing this as it has been superseded by the comprehensive core fixes in PR #243.

PR #427, opened 2026-03-20
fix(usenet): broadcast cond on Close() timeout to wake stuck goroutines

### LINKED ISSUE #426: Goroutine leak in UsenetReader.Close() — stuck downloadCond.Wait() after 30s timeout
### Description

When `Close()` times out waiting for in-flight downloads, it leaves two goroutines permanently leaked per reader instance:

1. The `downloadManager` goroutine stuck in `downloadCond.Wait()`
2. The anonymous close-waiter goroutine (`go func() { b.wg.Wait() }()`) inside `Close()` itself — this one can never complete since the download manager never exits

### Evidence

pprof goroutine dump shows (confirmed on a running instance):
- **26 goroutines** stuck in `downloadManager` at `usenet_reader.go` → `sync.Cond.Wait`
- **25 goroutines** stuck in `Close.func1.1` → `sync.WaitGroup.Wait`

These accumulate over time as files are opened and closed (e.g. Sonarr importing files that then fail).

### Root Cause

The `Close()` timeout branch (30s) cancels the context and clears `rg`, but never calls `b.downloadCond.Broadcast()`. The download manager is stuck in `Wait()` with no signal coming — the broadcast goroutine created at download start already fired when the context was cancelled, before `Wait()` was entered (missed signal).

### Fix

In the timeout branch of `Close()`, add a broadcast after clearing resources:

```go
case <-time.After(30 * time.Second):
    b.log.WarnContext(...)
    b.mu.Lock()
    if b.rg != nil {
        _ = b.rg.Clear()
        b.rg = nil
    }
    b.downloadCond.Broadcast() // wake stuck waiters so they can exit
    b.mu.Unlock()
```

Once woken, the download manager checks `ctx.Err()` (non-nil since `cancel()` was already called), exits the loop, and `pool.Wait()` finishes quickly since all download tasks have a cancelled context.

firestaerter3:
## Summary

- Adds `b.cond.Broadcast()` in the `Close()` 30s timeout branch
- Ensures any goroutines stuck in `cond.Wait()` are woken up after the timeout fires and `rg` is cleared

## Root Cause

When `Close()` times out waiting for in-flight downloads, goroutines stuck in `cond.Wait()` inside `downloadManager` were never woken up. The upfront `b.cond.Broadcast()` already handles the happy path (broadcast before the wait), but the timeout branch was missing it — leaving goroutines permanently leaked.

Confirmed via pprof on a live instance: 26 stuck `downloadManager` goroutines + 25 stuck close-waiter goroutines accumulating over time from Sonarr opening files that subsequently fail to import.

## Test plan

- [ ] Open and close files via FUSE/WebDAV with Sonarr triggering failed imports
- [ ] Check pprof goroutine count doesn't grow over time (`/debug/pprof/goroutine?debug=1`)
- [ ] Verify "Timeout waiting for downloads" log no longer results in permanently leaked goroutines

Closes #426

### DIFF
--- a/internal/usenet/usenet_reader.go
+++ b/internal/usenet/usenet_reader.go
@@ -161,7 +161,7 @@ func (b *UsenetReader) Close() error {
 				_ = b.rg.Clear()
 				b.rg = nil
 			}
-			b.downloadCond.Broadcast()
+			b.cond.Broadcast()
 			b.mu.Unlock()
 		}
 	})


javi11:
Thanks i end up doing it here https://github.com/javi11/altmount/pull/429

PR #448, opened 2026-03-26
feat: Allow for preserving internal subdir structure during NZB import

piefke84:
This PR contains changes to the file import and allows for preserving the file structure of nested RAR archives.

The code is not complete, parts missing:
- additional config var for "preserve_subdir_structure"
- implementation fpr 7z archives


javi11:
Thanks for this, i add it as default for both 7zip and rara https://github.com/javi11/altmount/pull/449

PR #483, opened 2026-04-11
FUSE: offset-native `ReadAt` reads, stable streaming prefetch, and tunable read-ahead

Defiance4331:
## Summary

Improves native **FUSE** playback with a single, coherent read path: kernel reads are served **offset-natively** via `ReadAt` / `ReadAtContext` instead of **`Seek` + `Read`**, avoiding repeated teardown of the forward Usenet reader that showed up as playback artifacts under aggressive read-ahead. **Sequential** `ReadAt` offsets reuse **one** streaming pipeline and prefetch; **non-sequential** offsets (probes, large backward jumps) still use **short-lived range readers** so correctness is preserved. **Hanwen** serializes `ReadAt` per open file handle so prefetch stays coherent under concurrent FUSE RPCs; **cgofuse** already serialized reads per handle and uses the same policy.

Adds **`fuse.use_read_at`** (default **on**) so operators can disable the offset-native path if needed, documents behavior in **streaming** docs, wires the option through **config API** and the **mount** UI, and updates the **default read-ahead** with tuning guidance.

## Motivation

Non-sequential kernel reads (read-ahead, index probes, container metadata) plus a **stateful** virtual file made **`Seek` + `Read`** expensive: each discontinuity reset the pipeline. A design that only opened a **new reader on every `ReadAt`** without reusing a sequential pipeline would fix the cursor issue but **starve prefetch** and hurt throughput. This PR delivers **both**: offset-native reads **and** a **shared sequential reader** on the hot path, with explicit per-handle ordering on hanwen.

## Why the default `max_read_ahead_mb` was lowered

FUSE read-ahead means the kernel can issue **read RPCs ahead of the app’s logical cursor**, sometimes far ahead. A **very large** window increases speculative reads at **many different offsets**, which stacks poorly with real clients that also do **probes** and **non-monotonic** access even during “normal” playback.

With an offset-native read path and **prefetch tied to a stable sequential pipeline**, a **more conservative default** (24 MiB in the updated defaults) **reduces far-ahead speculative churn** and keeps the workload closer to what prefetch can serve efficiently. The tradeoff is **somewhat more syscall/FUSE traffic** on very slow or very high-latency links.

Installs that previously relied on a **large** read-ahead (e.g. 128 MiB) for burst throughput can **raise `fuse.max_read_ahead_mb` again** after confirming stability—especially on high-latency paths where a bigger kernel window helps amortize round trips. The streaming docs describe this explicitly.

## Key changes

- **`internal/fuse/backend/hanwen/handle.go`**: `ReadAt` / `ReadAtContext` when enabled; per-handle ordering for the ReadAt path.
- **`internal/fuse/backend/cgofuse/fs.go`**: capability detection and ReadAt path where supported.
- **`internal/nzbfilesystem/metadata_remote_file.go`**: sequential `ReadAt` reuses the main reader; non-sequential fallback; WarmUp aligned with the path that serves bytes (see code + tests).
- **`internal/config/manager.go`**, OpenAPI/static swagger, **`MountConfigSection`**: **`fuse.use_read_at`**, read-ahead default updates, tests.
- **`docs/docs/3. Configuration/streaming.md`**: behavior, defaults, and tuning notes.

## Config / migration

- **`fuse.use_read_at`** (default **true**). Set **`false`** only as a fallback for an unexpected client; it restores the legacy read shape with known artifact tradeoffs on some workloads.
- **`fuse.max_read_ahead_mb`** reflects the new default in fresh defaults / UI; existing configs with an explicit value are unchanged. Operators who want the old throughput profile can set read-ahead back after validating playback.

## Testing

- Extended **`handle_test`** and **`metadata_remote_file`** coverage for ReadAt, interleaved offsets, and shared-path behavior.
- Please run **`make`** / project checks before merge (per repo convention).

---

**Disclaimer:** This pull request was produced with heavy AI assistance (“vibe-coded”).  **Please review the code, tests, and behavior carefully before merging.**

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -134,6 +134,7 @@ fuse:
     entry_timeout_seconds: 1
     max_cache_size_mb: 128
     max_read_ahead_mb: 128
+    use_read_at: true
 providers: []
 mount_path: ''
 profiler_enabled: false
--- a/docs/docs/3. Configuration/streaming.md
+++ b/docs/docs/3. Configuration/streaming.md
@@ -57,7 +57,9 @@ fuse:
   attr_timeout_seconds: 30
   entry_timeout_seconds: 1
   max_cache_size_mb: 128
-  max_read_ahead_mb: 128
+  max_read_ahead_mb: 24
+  # Prefer offset-native reads (default true). Set false only for debugging legacy Seek+Read behavior.
+  use_read_at: true
 
 segment_cache:
   enabled: true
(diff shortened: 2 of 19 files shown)


javi11:
Fixed here https://github.com/javi11/altmount/pull/484 thanks for the help

PR #517, opened 2026-04-19
feat(importer): compress NZBs to .nzb.gz in persistent storage

javi11:
## Summary

- NZB files (XML) are now stored gzip-compressed as `.nzb.gz` in the persistent `.nzbs/` directory, saving ~80-90% disk space
- Added `openNzbFile` helper that transparently decompresses `.nzb.gz` or passes through plain `.nzb` files, ensuring backward compatibility with existing uncompressed NZBs
- All read paths (`ProcessNzbFile`, `CalculateFileSizeOnly`) now use `openNzbFile`
- `RegenerateMetadata` updated to find both `.nzb` and `.nzb.gz` files when searching `.nzbs/`

## What changed

**New file: `internal/importer/nzb_file.go`**
- `openNzbFile(path)` — opens `.nzb` or `.nzb.gz` returning an `io.ReadCloser`
- `compressNzbToGz(src, dst)` — writes a gzip-compressed copy of an NZB

**`internal/importer/service.go`**
- `ensurePersistentNzb`: destination filename now always ends in `.nzb.gz`; uses rename+compress (same FS) or direct compress (cross-device) instead of plain copy
- `CalculateFileSizeOnly`: uses `openNzbFile` for NZB paths
- `RegenerateMetadata`: WalkDir callback now matches both `.nzb` and `.nzb.gz`

**`internal/importer/processor.go`**
- `ProcessNzbFile`: uses `openNzbFile` instead of `os.Open`; removed now-unused `os` import

## Test plan
- [ ] Unit tests in `nzb_file_test.go` cover plain `.nzb`, `.nzb.gz`, and compress round-trip
- [ ] All importer tests pass with `-race`: `go test ./internal/importer/... -race`
- [ ] Full `make` build passes
- [ ] Existing `.nzb` files in `.nzbs/` still process correctly (backward compat via `openNzbFile`)

### DIFF
--- a/internal/api/queue_handlers.go
+++ b/internal/api/queue_handlers.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"compress/gzip"
 	"fmt"
 	"html"
 	"io"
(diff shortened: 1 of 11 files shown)


lostgit:
This is great. Will existing files .nzbs get compressed? If not, is it safe to manually run gzip on them? (once the new version is running)

javi11:
Yes they will atomatically be compressed

PR #524, opened 2026-04-21
fix(importer): correctly strip .nzb and .nzb.gz extensions from release names

drondeseries:
### Description
This PR fixes an issue where the `.nzb` extension was being included in virtual directory names and symlink filenames (e.g., `Movie.nzb/Movie.nzb.mkv`).

### Root Cause
The issue was introduced in commit `23bfde95` when transparent NZB compression (`.nzb.gz`) was added to persistent storage. The existing codebase relied on `filepath.Ext()` to strip extensions. However, `filepath.Ext()` only identifies the final extension (`.gz`), leaving the `.nzb` part intact.

### Changes
- Introduced a centralized, dependency-free utility package `internal/importer/utils/nzbtrim`.
- Implemented `TrimNzbExtension` to handle case-insensitive removal of both `.nzb` and `.nzb.gz`.
- Refactored all occurrences of `filepath.Ext` and `strings.TrimSuffix(..., ".nzb")` in the import and API logic to use this new utility.

### Impacted Areas
- **Virtual File System**: Folder naming in `internal/importer/filesystem/utils.go`.
- **Processors**: Single-file, multi-file, RAR, and 7zip release name derivation.
- **API**: Stremio handler (`nzb_stremio_handlers.go`) now generates clean stream names.
- **Service**: Persistence and metadata regeneration logic in `service.go`.

### Verification
- Verified with reproduction tests ensuring `file.nzb.gz` and `file.NZB` both resolve to `file`.
- Confirmed no regression in existing importer test suites.

### DIFF
--- a/internal/api/arrs_handlers.go
+++ b/internal/api/arrs_handlers.go
@@ -337,17 +337,23 @@ func (s *Server) handleArrsWebhook(c *fiber.Ctx) error {
 		}
 
 		// Redundant Deletion Guard: ensure the file is gone from the local mount
-		if s.configManager != nil {
+		if s.configManager != nil && metadataPath != "" && metadataPath != "." && metadataPath != "/" {
 			cfg := s.configManager.GetConfig()
 			if cfg.MountPath != "" {
 				localPath := filepath.Join(cfg.MountPath, metadataPath)
+
+				// HARD SAFETY: Never delete the mount root or critical system paths
+				cleanLocal := filepath.Clean(localPath)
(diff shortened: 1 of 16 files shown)


lostgit:
altmount supports password-protected 7zip, right? I would think so but thought I'd confirm because was getting a few import error saying unable to open password-protected 7zip, but I think it's the articles themselves that are the issue.

javi11 (approved):

PR #541, opened 2026-04-27
fix(streaming): coalesce ARR-repair triggers and force-unmount FUSE on recovery exhaustion

### LINKED ISSUE #3: saving configuration
After the server restart all the configuration settings gets reset.

### LINKED ISSUE #4: altmount completion bash Error
running `altmount completion bash` gives the following error:

```
# bash completion V2 for altmount                             -*- shell-script -*-

__altmount_debug()
{
    if [[ -n ${BASH_COMP_DEBUG_FILE-} ]]; then
        echo "$*" >> "${BASH_COMP_DEBUG_FILE}"
    fi
}

# Macs have bash3 for which the bash-completion package doesn't include
# _init_completion. This is a minimal version of that function.
__altmount_init_completion()
{
    COMPREPLY=()
    _get_comp_words_by_ref "$@" cur prev words cword
}

# This function calls the altmount program to obtain the completion
# results and the directive.  It fills the 'out' and 'directive' vars.
__altmount_get_completion_results() {
    local requestComp lastParam lastChar args

    # Prepare the command to request completions for the program.
    # Calling ${words[0]} instead of directly altmount allows handling aliases
    args=("${words[@]:1}")
    requestComp="${words[0]} __complete ${args[*]}"

    lastParam=${words[$((${#words[@]}-1))]}
    lastChar=${lastParam:$((${#lastParam}-1)):1}
    __altmount_debug "lastParam ${lastParam}, lastChar ${lastChar}"

    if [[ -z ${cur} && ${lastChar} != = ]]; then
        # If the last parameter is complete (there is a space following it)
        # We add an extra empty parameter so we can indicate this to the go method.
        __altmount_debug "Adding extra empty parameter"
        requestComp="${requestComp} ''"
    fi

    # When completing a flag with an = (e.g., altmount -n=<TAB>)
    # bash focuses on the part after the =, so we need to remove
    # the flag part from $cur
    if [[ ${cur} == -*=* ]]; then
        cur="${cur#*=}"
    fi

    __altmount_debug "Calling ${requestComp}"
    # Use eval to handle any environment variables and such
    out=$(eval "${requestComp}" 2>/dev/null)

    # Extract the directive integer at the very end of the output following a colon (:)
    directive=${out##*:}
    # Remove the directive
    out=${out%:*}
    if [[ ${directive} == "${out}" ]]; then
        # There is not directive specified
        directive=0
    fi
    __altmount_debug "The completion directive is: ${directive}"
    __altmount_debug "The completions are: ${out}"
}

__altmount_process_completion_results() {
    local shellCompDirectiveError=1
    local shellCompDirectiveNoSpace=2
    local shellCompDirectiveNoFileComp=4
    local shellCompDirectiveFilterFileExt=8
    local shellCompDirectiveFilterDirs=16
    local shellCompDirectiveKeepOrder=32

    if (((directive & shellCompDirectiveError) != 0)); then
        # Error code.  No completion.
        __altmount_debug "Received error from custom completion go code"
        return
    else
        if (((directive & shellCompDirectiveNoSpace) != 0)); then
            if [[ $(type -t compopt) == builtin ]]; then
                __altmount_debug "Activating no space"
                compopt -o nospace
            else
                __altmount_debug "No space directive not supported in this version of bash"
            fi
        fi
        if (((directive & shellCompDirectiveKeepOrder) != 0)); then
            if [[ $(type -t compopt) == builtin ]]; then
                # no sort isn't supported for bash less than < 4.4
                if [[ ${BASH_VERSINFO[0]} -lt 4 || ( ${BASH_VERSINFO[0]} -eq 4 && ${BASH_VERSINFO[1]} -lt 4 ) ]]; then
                    __altmount_debug "No sort directive not supported in this version of bash"
                else
                    __altmount_debug "Activating keep order"
                    compopt -o nosort
                fi
            else
                __altmount_debug "No sort directive not supported in this version of bash"
            fi
        fi
        if (((directive & shellCompDirectiveNoFileComp) != 0)); then
            if [[ $(type -t compopt) == builtin ]]; then
                __altmount_debug "Activating no file completion"
                compopt +o default
            else
                __altmount_debug "No file completion directive not supported in this version of bash"
            fi
        fi
    fi

    # Separate activeHelp from normal completions
    local completions=()
    local activeHelp=()
    __altmount_extract_activeHelp

    if (((directive & shellCompDirectiveFilterFileExt) != 0)); then
        # File extension filtering
        local fullFilter="" filter filteringCmd

        # Do not use quotes around the $completions variable or else newline
        # characters will be kept.
        for filter in ${completions[*]}; do
            fullFilter+="$filter|"
        done

        filteringCmd="_filedir $fullFilter"
        __altmount_debug "File filtering command: $filteringCmd"
        $filteringCmd
    elif (((directive & shellCompDirectiveFilterDirs) != 0)); then
        # File completion for directories only

        local subdir
        subdir=${completions[0]}
        if [[ -n $subdir ]]; then
            __altmount_debug "Listing directories in $subdir"
            pushd "$subdir" >/dev/null 2>&1 && _filedir -d && popd >/dev/null 2>&1 || return
        else
            __altmount_debug "Listing directories in ."
            _filedir -d
        fi
    else
        __altmount_handle_completion_types
    fi

    __altmount_handle_special_char "$cur" :
    __altmount_handle_special_char "$cur" =

    # Print the activeHelp statements before we finish
    __altmount_handle_activeHelp
}

__altmount_handle_activeHelp() {
    # Print the activeHelp statements
    if ((${#activeHelp[*]} != 0)); then
        if [ -z $COMP_TYPE ]; then
            # Bash v3 does not set the COMP_TYPE variable.
            printf "\n";
            printf "%s\n" "${activeHelp[@]}"
            printf "\n"
            __altmount_reprint_commandLine
            return
        fi

        # Only print ActiveHelp on the second TAB press
        if [ $COMP_TYPE -eq 63 ]; then
            printf "\n"
            printf "%s\n" "${activeHelp[@]}"

            if ((${#COMPREPLY[*]} == 0)); then
                # When there are no completion choices from the program, file completion
                # may kick in if the program has not disabled it; in such a case, we want
                # to know if any files will match what the user typed, so that we know if
                # there will be completions presented, so that we know how to handle ActiveHelp.
                # To find out, we actually trigger the file completion ourselves;
                # the call to _filedir will fill COMPREPLY if files match.
                if (((directive & shellCompDirectiveNoFileComp) == 0)); then
                    __altmount_debug "Listing files"
                    _filedir
                fi
            fi

            if ((${#COMPREPLY[*]} != 0)); then
                # If there are completion choices to be shown, print a delimiter.
                # Re-printing the command-line will automatically be done
                # by the shell when it prints the completion choices.
                printf -- "--"
            else
                # When there are no completion choices at all, we need
                # to re-print the command-line since the shell will
                # not be doing it itself.
                __altmount_reprint_commandLine
            fi
        elif [ $COMP_TYPE -eq 37 ] || [ $COMP_TYPE -eq 42 ]; then
            # For completion type: menu-complete/menu-complete-backward and insert-completions
            # the completions are immediately inserted into the command-line, so we first
            # print the activeHelp message and reprint the command-line since the shell won't.
            printf "\n"
            printf "%s\n" "${activeHelp[@]}"

            __altmount_reprint_commandLine
        fi
    fi
}

__altmount_reprint_commandLine() {
    # The prompt format is only available from bash 4.4.
    # We test if it is available before using it.
    if (x=${PS1@P}) 2> /dev/null; then
        printf "%s" "${PS1@P}${COMP_LINE[@]}"
    else
        # Can't print the prompt.  Just print the
        # text the user had typed, it is workable enough.
        printf "%s" "${COMP_LINE[@]}"
    fi
}

# Separate activeHelp lines from real completions.
# Fills the $activeHelp and $completions arrays.
__altmount_extract_activeHelp() {
    local activeHelpMarker="_activeHelp_ "
    local endIndex=${#activeHelpMarker}

    while IFS='' read -r comp; do
        [[ -z $comp ]] && continue

        if [[ ${comp:0:endIndex} == $activeHelpMarker ]]; then
            comp=${comp:endIndex}
            __altmount_debug "ActiveHelp found: $comp"
            if [[ -n $comp ]]; then
                activeHelp+=("$comp")
            fi
        else
            # Not an activeHelp line but a normal completion
            completions+=("$comp")
        fi
    done <<<"${out}"
}

__altmount_handle_completion_types() {
    __altmount_debug "__altmount_handle_completion_types: COMP_TYPE is $COMP_TYPE"

    case $COMP_TYPE in
    37|42)
        # Type: menu-complete/menu-complete-backward and insert-completions
        # If the user requested inserting one completion at a time, or all
        # completions at once on the command-line we must remove the descriptions.
        # https://github.com/spf13/cobra/issues/1508

        # If there are no completions, we don't need to do anything
        (( ${#completions[@]} == 0 )) && return 0

        local tab=$'\t'

        # Strip any description and escape the completion to handled special characters
        IFS=$'\n' read -ra completions -d '' < <(printf "%q\n" "${completions[@]%%$tab*}")

        # Only consider the completions that match
        IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n'; compgen -W "${completions[*]}" -- "${cur}")

        # compgen looses the escaping so we need to escape all completions again since they will
        # all be inserted on the command-line.
        IFS=$'\n' read -ra COMPREPLY -d '' < <(printf "%q\n" "${COMPREPLY[@]}")
        ;;

    *)
        # Type: complete (normal completion)
        __altmount_handle_standard_completion_case
        ;;
    esac
}

__altmount_handle_standard_completion_case() {
    local tab=$'\t'

    # If there are no completions, we don't need to do anything
    (( ${#completions[@]} == 0 )) && return 0

    # Short circuit to optimize if we don't have descriptions
    if [[ "${completions[*]}" != *$tab* ]]; then
        # First, escape the completions to handle special characters
        IFS=$'\n' read -ra completions -d '' < <(printf "%q\n" "${completions[@]}")
        # Only consider the completions that match what the user typed
        IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n'; compgen -W "${completions[*]}" -- "${cur}")

        # compgen looses the escaping so, if there is only a single completion, we need to
        # escape it again because it will be inserted on the command-line.  If there are multiple
        # completions, we don't want to escape them because they will be printed in a list
        # and we don't want to show escape characters in that list.
        if (( ${#COMPREPLY[@]} == 1 )); then
            COMPREPLY[0]=$(printf "%q" "${COMPREPLY[0]}")
        fi
        return 0
    fi

    local longest=0
    local compline
    # Look for the longest completion so that we can format things nicely
    while IFS='' read -r compline; do
        [[ -z $compline ]] && continue

        # Before checking if the completion matches what the user typed,
        # we need to strip any description and escape the completion to handle special
        # characters because those escape characters are part of what the user typed.
        # Don't call "printf" in a sub-shell because it will be much slower
        # since we are in a loop.
        printf -v comp "%q" "${compline%%$tab*}" &>/dev/null || comp=$(printf "%q" "${compline%%$tab*}")

        # Only consider the completions that match
        [[ $comp == "$cur"* ]] || continue

        # The completions matches.  Add it to the list of full completions including
        # its description.  We don't escape the completion because it may get printed
        # in a list if there are more than one and we don't want show escape characters
        # in that list.
        COMPREPLY+=("$compline")

        # Strip any description before checking the length, and again, don't escape
        # the completion because this length is only used when printing the completions
        # in a list and we don't want show escape characters in that list.
        comp=${compline%%$tab*}
        if ((${#comp}>longest)); then
            longest=${#comp}
        fi
    done < <(printf "%s\n" "${completions[@]}")

    # If there is a single completion left, remove the description text and escape any special characters
    if ((${#COMPREPLY[*]} == 1)); then
        __altmount_debug "COMPREPLY[0]: ${COMPREPLY[0]}"
        COMPREPLY[0]=$(printf "%q" "${COMPREPLY[0]%%$tab*}")
        __altmount_debug "Removed description from single completion, which is now: ${COMPREPLY[0]}"
    else
        # Format the descriptions
        __altmount_format_comp_descriptions $longest
    fi
}

__altmount_handle_special_char()
{
    local comp="$1"
    local char=$2
    if [[ "$comp" == *${char}* && "$COMP_WORDBREAKS" == *${char}* ]]; then
        local word=${comp%"${comp##*${char}}"}
        local idx=${#COMPREPLY[*]}
        while ((--idx >= 0)); do
            COMPREPLY[idx]=${COMPREPLY[idx]#"$word"}
        done
    fi
}

__altmount_format_comp_descriptions()
{
    local tab=$'\t'
    local comp desc maxdesclength
    local longest=$1

    local i ci
    for ci in ${!COMPREPLY[*]}; do
        comp=${COMPREPLY[ci]}
        # Properly format the description string which follows a tab character if there is one
        if [[ "$comp" == *$tab* ]]; then
            __altmount_debug "Original comp: $comp"
            desc=${comp#*$tab}
            comp=${comp%%$tab*}

            # $COLUMNS stores the current shell width.
            # Remove an extra 4 because we add 2 spaces and 2 parentheses.
            maxdesclength=$(( COLUMNS - longest - 4 ))

            # Make sure we can fit a description of at least 8 characters
            # if we are to align the descriptions.
            if ((maxdesclength > 8)); then
                # Add the proper number of spaces to align the descriptions
                for ((i = ${#comp} ; i < longest ; i++)); do
                    comp+=" "
                done
            else
                # Don't pad the descriptions so we can fit more text after the completion
                maxdesclength=$(( COLUMNS - ${#comp} - 4 ))
            fi

            # If there is enough space for any description text,
            # truncate the descriptions that are too long for the shell width
            if ((maxdesclength > 0)); then
                if ((${#desc} > maxdesclength)); then
                    desc=${desc:0:$(( maxdesclength - 1 ))}
                    desc+="…"
                fi
                comp+="  ($desc)"
            fi
            COMPREPLY[ci]=$comp
            __altmount_debug "Final comp: $comp"
        fi
    done
}

__start_altmount()
{
    local cur prev words cword split

    COMPREPLY=()

    # Call _init_completion from the bash-completion package
    # to prepare the arguments properly
    if declare -F _init_completion >/dev/null 2>&1; then
        _init_completion -n =: || return
    else
        __altmount_init_completion -n =: || return
    fi

    __altmount_debug
    __altmount_debug "========= starting completion logic =========="
    __altmount_debug "cur is ${cur}, words[*] is ${words[*]}, #words[@] is ${#words[@]}, cword is $cword"

    # The user could have moved the cursor backwards on the command-line.
    # We need to trigger completion from the $cword location, so we need
    # to truncate the command-line ($words) up to the $cword location.
    words=("${words[@]:0:$cword+1}")
    __altmount_debug "Truncated words[*]: ${words[*]},"

    local out directive
    __altmount_get_completion_results
    __altmount_process_completion_results
}

if [[ $(type -t compopt) = "builtin" ]]; then
    complete -o default -F __start_altmount altmount
else
    complete -o default -o nospace -F __start_altmount altmount
fi

# ex: ts=4 sw=4 et filetype=sh
```

### LINKED ISSUE #539: v0.2.0: NNTP 430 batch triggers ARR-repair storm, then mount recovery fails after 4 attempts and leaves ghost mount
## Summary

When altmount encounters a batch of NZB files where the underlying Usenet
articles are gone (`nntp: 430 no such article`), it fires an "immediate ARR
repair" for each one in rapid succession. During or right after that storm,
the rclone RC API at `localhost:5573` becomes unresponsive, mount recovery
fails after 4 attempts, and the FUSE mount at `/mnt/altmount` is left in a
"ghost" state: kernel still considers it mounted, but every read blocks
indefinitely. The container's own logs say "Failed to recover mount" but the
container keeps reporting healthy.

The hung mount stays broken until the container (or host) is restarted.
Meanwhile any process touching `/mnt/altmount` accumulates in D-state. In our
case ~16 hours later the LXC load average reached 600 and dockerd became
unresponsive.

## Version

- Image: `ghcr.io/javi11/altmount:latest`
- Built: 2026-04-14
- Labels: `org.opencontainers.image.version=0.2.0-amd64-temp`,
  `revision=dc80b5bbda0d51c7c704e5f5f956cec043ccfe06`

## Setup

- Debian 12 LXC on Proxmox
- altmount mount used by Sonarr, Radarr, Plex, Jellyfin
- ~12 episodes of a TV series queued, all with expired Usenet articles

## Timeline (from container logs, internal timestamps)

```
20:54:27   File corrupted   "nntp: 430 no such article"   episode 1
20:54:27   Streaming failure detected, triggering immediate ARR repair
20:54:27   Moving metadata file for corrupted item to safety folder
20:55:18   File corrupted ... episode 3
20:55:47   File corrupted ... episode 4
20:56:??   ... episodes 5..11 ...
20:58:02   File corrupted ... episode 12
20:57:30   Creating mount directory                                   (recovery attempt 1)
21:04:11   "Nothing to unmount" / all force unmount attempts failed
21:05:03   Mount attempt failed ... attempt=4
           err="failed to create rclone config: failed to create config altmount:
                failed to make request:
                Post http://localhost:5573/config/create:
                context deadline exceeded (Client.Timeout exceeded while awaiting headers)"
21:05:32   Failed to recover mount for altmount: mount failed for altmount
```

After 21:05 there are no further altmount log entries about the mount, but
`mount | grep altmount` still showed the mount, and `ls /mnt/altmount` from
host or any consumer container blocked indefinitely. Sonarr/Radarr ffprobe
processes scanning files under that path piled up in D-state.

## Suspected pattern

Looks related to #70 (alpha5 - high cpu usage randomly - freezing mount),
#75 (Failed hangs after 3 retries), and #127 (Proper unmount when stop
almount), all closed but the symptom is reproducing on v0.2.0.

The "context deadline exceeded" on `Post /config/create` suggests the rclone
RC daemon inside the container had stopped responding by the time recovery
ran. That fits a scenario where the in-flight ARR-repair traffic and the
recovery loop are competing for the same rclone process.

## What would help

1. Cap parallelism on "immediate ARR repair" so a corrupt batch does not
   fire N triggers simultaneously (queue with a small concurrency limit).
2. Make mount recovery more aggressive about detecting and bailing on a
   wedged rclone RC: kill the rclone process and respawn rather than loop
   on `/config/create` with the same hung backend.
3. On final "Failed to recover mount", also `fusermount -uz` the stale
   mountpoint so consumers stop blocking on dead reads (currently the
   ghost mount persists and kills the host).
4. Container healthcheck could include a `wget -T3 :5573/` probe so the
   container is reported unhealthy when the RC daemon is wedged.

Happy to provide more logs or run instrumented builds if useful.

javi11:
Closes #539.

## Problem

When a batch of NZBs has expired Usenet articles (`nntp: 430 no such article`), every streaming failure synchronously fired an "immediate ARR repair" plus an async `vfs/refresh` POST to the rclone RC daemon — one per file, with no debounce. With ~12 corrupted files this storms the RC daemon at `localhost:5573`. The FUSE health monitor then tries to recover and fails after 3 attempts on `Post /config/create: context deadline exceeded`. After exhaustion, `recoverMount` simply `return`s — the kernel still holds the mount, every read on `/mnt/altmount` blocks indefinitely, consumer processes (Sonarr/Radarr/Plex/ffprobe) pile up in D-state, and the container keeps reporting healthy. In the reporter's environment the host load average reached 600 after ~16h.

### 1. Storm throttling — `internal/nzbfilesystem/repair_coalescer.go` (new)

`RepairCoalescer` provides two mechanisms shared across all `MetadataVirtualFile` instances:

- **Per-path debounce (30s window).** `ShouldTrigger(path)` returns `true` at most once per debounce window per path; repeat reads of the same corrupted file no longer re-fire the repair pipeline.
- **Directory-level VFS-refresh coalescing.** `EnqueueRefresh(virtualDir)` adds the directory to a pending set; a single worker goroutine drains the set after a 1s flush window, dedupes, and issues at most one `RefreshDir` RC call at a time with all affected dirs batched into one request.

Result: a burst of 12 concurrent NNTP 430s in the same dir → 1 RC POST instead of 12.

Wired into `MetadataRemoteFile` via `NewRepairCoalescer` (no public signature change) and propagated to each `MetadataVirtualFile` on `OpenFile`. `updateFileHealthOnError` short-circuits via the debounce and replaces the per-file `RefreshDir` goroutine with a coalescer enqueue.

### 2. Force-unmount on FUSE recovery exhaustion — `internal/api/fuse_handlers.go`

`FuseManager.recoverMount` previously logged and `return`ed on the exhaustion path, leaving a kernel-side ghost mount. It now calls `server.ForceUnmount()` (which runs `fusermount(3) -uz` / `umount -lf` under the hood) before returning, and transitions to a new `"failed"` terminal state so future status calls reflect the dead mount.

## Tests

`internal/nzbfilesystem/repair_coalescer_test.go` — 5 race-clean tests:

- `ShouldTrigger_DebouncesByPath` — same path debounced; different paths independent; TTL expiry works.
- `ShouldTrigger_NilReceiverIsSafe` — nil-receiver no-op fallback for tests.
- `BurstCoalescesIntoSingleRefresh` — 12 concurrent enqueues for one dir collapse to exactly 1 RC call.
- `DistinctDirsBatchTogether` — 3 distinct dirs + 1 duplicate → 1 batched call with 3 deduped entries.
- `NilClient_NoPanic` — coalescer with nil client doesn't panic.

`go test -race -count=1 ./internal/nzbfilesystem ./internal/api` — pass.
`go vet ./...` — clean. Full `go build ./...` — clean.

## Reviewer notes

- This PR addresses the two highest-blast-radius items in the issue (the storm trigger and the host-killing ghost mount). Two follow-ups remain from the report and can ship in subsequent PRs:
  - Kill+respawn the rclone process during recovery if its RC has wedged (fix #3 in the plan).
  - Add an RC liveness probe to `/api/health` so Docker `HEALTHCHECK` reflects a wedged daemon (fix #4).
- Backwards compatibility: `MetadataVirtualFile.repairCoalescer` is nil-safe, so existing test fixtures that build the struct directly continue to compile and run.
- Lifecycle: the coalescer's worker goroutine has process lifetime; `Close()` is provided for test cleanup.

## Test plan

- [x] Unit tests pass with `-race`.
- [x] `go vet` and `go build` clean.
- [ ] E2E reproduction in reporter's scenario (queue ~12 NZBs with missing articles, confirm logs show coalesced refreshes and no ghost mount on simulated RC wedge — see plan's verification section).

### DIFF
--- a/internal/api/fuse_handlers.go
+++ b/internal/api/fuse_handlers.go
@@ -123,9 +123,27 @@ func (m *FuseManager) recoverMount(ctx context.Context) {
 	m.mu.Unlock()
 
 	if attempt > fuseMaxRecoveryAttempts {
-		slog.ErrorContext(ctx, "FUSE recovery exhausted, staying in error state",
+		// Recovery has failed. Detach the kernel-side FUSE mount so consumers
+		// (Sonarr/Radarr/Plex/ffprobe etc.) stop blocking indefinitely on dead
+		// reads — leaving a "ghost" mount accumulates D-state processes and can
+		// take down the host (see issue #539). ForceUnmount runs
+		// `fusermount(3) -uz` / `umount -lf` under the hood.
+		slog.ErrorContext(ctx, "FUSE recovery exhausted, force-unmounting to release kernel mount",
 			"attempts", attempt-1,
 			"path", path)
+
+		if server != nil {
+			if err := server.ForceUnmount(); err != nil {
+				slog.ErrorContext(ctx, "Force unmount on recovery exhaustion failed",
+					"path", path,
+					"error", err)
+			}
+		}
+
+		m.mu.Lock()
+		m.server = nil
+		m.status = "failed"
+		m.mu.Unlock()
 		return
 	}
 
--- a/internal/nzbfilesystem/metadata_remote_file.go
+++ b/internal/nzbfilesystem/metadata_remote_file.go
@@ -41,6 +41,7 @@ type MetadataRemoteFile struct {
 	aesCipher        *aes.AesCipher           // For AES encryption/decryption
 	streamTracker    StreamTracker            // Stream tracker for monitoring active streams
 	cacheSource      *segcache.Source         // Segment cache source (nil = no cache configured)
+	repairCoalescer  *RepairCoalescer         // Throttles streaming-failure repair triggers and rclone VFS refreshes
 	renameMu         sync.Mutex               // Mutex to protect rename operations from race conditions
 }
 
@@ -81,6 +82,7 @@ func NewMetadataRemoteFile(
 		aesCipher:        aesCipher,
 		streamTracker:    streamTracker,
 		cacheSource:      cacheSource,
+		repairCoalescer:  NewRepairCoalescer(rcloneClient, configGetter),
 	}
 }
 
@@ -263,6 +265,7 @@ func (mrf *MetadataRemoteFile) OpenFile(ctx context.Context, name string) (bool,
 		healthRepository: mrf.healthRepository,
 		arrsService:      mrf.arrsService,
 		rcloneClient:     mrf.rcloneClient,
+		repairCoalescer:  mrf.repairCoalescer,
 		configGetter:     mrf.configGetter,
 		poolManager:      mrf.poolManager,
 		ctx:              ctx,
@@ -773,6 +776,7 @@ type MetadataVirtualFile struct {
 	healthRepository *database.HealthRepository
 	arrsService      ARRsRepairService
 	rcloneClient     rclonecli.RcloneRcClient // RClone RC client for VFS notifications
+	repairCoalescer  *RepairCoalescer         // Throttles repair triggers; may be nil in tests
 	configGetter     config.ConfigGetter
 	poolManager      pool.Manager // Pool manager for dynamic pool access
 	ctx              context.Context
@@ -1688,7 +1692,21 @@ func (mvf *MetadataVirtualFile) wrapWithEncryption(start, end int64) (io.ReadClo
 
 // updateFileHealthOnError updates both metadata and database health status when corruption is detected.
 // Uses synchronous operations with timeout to prevent goroutine leaks.
+//
+// A streaming-failure repair trigger for the same path is debounced through the
+// shared RepairCoalescer so that repeated corrupt reads of one file (or a batch
+// of corrupt files) cannot fan out into one DB write + one rclone VFS refresh
+// per call. See issue #539 for the failure mode this guards against.
 func (mvf *MetadataVirtualFile) updateFileHealthOnError(dataCorruptionErr *usenet.DataCorruptionError, noRetry bool) {
+	// Per-path debounce: short-circuit if this file already triggered a repair
+	// inside the debounce window. ShouldTrigger handles a nil coalescer
+	// (test harness) by returning true.
+	if !mvf.repairCoalescer.ShouldTrigger(mvf.name) {
+		slog.DebugContext(mvf.ctx, "Streaming failure repair already triggered recently, debouncing",
+			"file", mvf.name)
+		return
+	}
+
 	// Use a short timeout context to prevent blocking indefinitely
 	ctx, cancel := context.WithTimeout(mvf.ctx, 5*time.Second)
 	defer cancel()
@@ -1727,22 +1745,12 @@ func (mvf *MetadataVirtualFile) updateFileHealthOnError(dataCorruptionErr *usene
 			relativePath = strings.TrimPrefix(relativePath, "/")
 			slog.InfoContext(ctx, "Moving metadata file for corrupted item to safety folder to trigger replacement", "file_path", mvf.name)
 			if moveErr := mvf.metadataService.MoveToCorrupted(ctx, relativePath); moveErr == nil {
-				// Successfully moved metadata, notify rclone VFS refresh
-				if mvf.rcloneClient != nil {
-					go func() {
-						virtualDir := filepath.Dir(mvf.name)
-						vfsName := cfg.RClone.VFSName
-						if vfsName == "" {
-							vfsName = config.MountProvider
-						}
-						// Increased timeout to 60 seconds as vfs/refresh can be slow
-						vfsCtx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
-						defer cancel()
-						if refreshErr := mvf.rcloneClient.RefreshDir(vfsCtx, vfsName, []string{virtualDir}); refreshErr != nil {
-							slog.ErrorContext(vfsCtx, "Failed to notify rclone VFS about file status change after streaming failure", "file", mvf.name, "err", refreshErr)
-						}
-					}()
-				}
+				// Successfully moved metadata, enqueue a coalesced rclone VFS
+				// refresh. Multiple files in the same directory collapse into a
+				// single RC call; concurrent failures across directories are
+				// batched into one call as well. EnqueueRefresh is a no-op on a
+				// nil coalescer (test harness).
+				mvf.repairCoalescer.EnqueueRefresh(filepath.Dir(mvf.name))
 			} else {
 				slog.WarnContext(ctx, "Failed to move corrupted metadata file, proceeding with repair trigger status", "error", moveErr)
 			}
--- a/internal/nzbfilesystem/repair_coalescer.go
+++ b/internal/nzbfilesystem/repair_coalescer.go
@@ -1,0 +1,181 @@
+package nzbfilesystem
+
+import (
+	"context"
+	"log/slog"
+	"sync"
+	"time"
+
+	"github.com/javi11/altmount/internal/config"
+	"github.com/javi11/altmount/pkg/rclonecli"
+)
+
+// RepairCoalescer throttles per-file streaming-failure repair triggers and
+// coalesces rclone VFS refresh calls so that a batch of corrupted reads cannot
+// fan out into one RC POST per file.
+//
+// Two mechanisms:
+//
+//  1. Per-path debounce — ShouldTrigger returns false if the same file path
+//     was triggered within the debounce window. This prevents repeated reads
+//     of the same corrupted file from re-firing the repair pipeline.
+//
+//  2. Directory-level VFS refresh coalescing — EnqueueRefresh adds a virtual
+//     directory to a pending set; a single worker goroutine drains the set,
+//     deduplicates, and issues at most one RefreshDir RC call at a time with
+//     a minimum interval between calls.
+type RepairCoalescer struct {
+	rclone       rclonecli.RcloneRcClient
+	configGetter config.ConfigGetter
+
+	debounceTTL  time.Duration
+	flushDelay   time.Duration
+	refreshTO    time.Duration
+
+	mu      sync.Mutex
+	seen    map[string]time.Time
+	pending map[string]struct{}
+
+	wakeCh chan struct{}
+	stopCh chan struct{}
+	stopWg sync.WaitGroup
+}
+
+const (
+	defaultRepairDebounceTTL = 30 * time.Second
+	defaultRefreshFlushDelay = 1 * time.Second
+	defaultRefreshTimeout    = 60 * time.Second
+)
+
+// NewRepairCoalescer constructs a coalescer and starts its background worker.
+// The worker runs for the lifetime of the process; call Close to stop it in
+// tests or during graceful shutdown.
+func NewRepairCoalescer(rclone rclonecli.RcloneRcClient, configGetter config.ConfigGetter) *RepairCoalescer {
+	c := &RepairCoalescer{
+		rclone:       rclone,
+		configGetter: configGetter,
+		debounceTTL:  defaultRepairDebounceTTL,
+		flushDelay:   defaultRefreshFlushDelay,
+		refreshTO:    defaultRefreshTimeout,
+		seen:         make(map[string]time.Time),
+		pending:      make(map[string]struct{}),
+		wakeCh:       make(chan struct{}, 1),
+		stopCh:       make(chan struct{}),
+	}
+	c.stopWg.Add(1)
+	go c.run()
+	return c
+}
+
+// ShouldTrigger reports whether a streaming-failure repair should be fired for
+// path. It returns true at most once per debounce window per path. Stale
+// entries are pruned opportunistically on each call.
+func (c *RepairCoalescer) ShouldTrigger(path string) bool {
+	if c == nil {
+		return true
+	}
+	now := time.Now()
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if last, ok := c.seen[path]; ok && now.Sub(last) < c.debounceTTL {
+		return false
+	}
+	c.seen[path] = now
+
+	// Opportunistic prune: cap the map growth on long-running processes.
+	if len(c.seen) > 1024 {
+		cutoff := now.Add(-c.debounceTTL)
+		for k, t := range c.seen {
+			if t.Before(cutoff) {
+				delete(c.seen, k)
+			}
+		}
+	}
+	return true
+}
+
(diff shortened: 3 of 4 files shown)


firestaerter3:
Thanks Javi!

PR #601, opened 2026-05-20
fix: connection-lifecycle

iPromKnight:
Connection-lifecycle storms (S1/S3/S5/S6/S8/S9)
Re-opening of [#589](https://github.com/javi11/altmount/issues/589) (branch was accidentally deleted after merge to dev).

Scope is connection-lifecycle and observability only. Cross-process budgeting between streams and imports is out of scope — that lives in [#595](https://github.com/javi11/altmount/issues/595).

Fixes

S1 / S3 — retry.Attempts(5) → Attempts(2) with [50, 150]ms jitter, so a permanently failing segment caps at 2 wire calls and concurrent retriers desync against a recovering provider instead of thundering-herd.
S5 — per-file segment LRU on the ephemeral ReadAt path coalesces Plex/Jellyfin scrub bursts (200 small reads in an 8-segment window → 8 wire calls). Plus an EOF-straddle slice-bound fix that was panicking on partially-filled final segments.
S6 — bounded closer-worker pool absorbs seek-spam (50-seek burst → ~4 goroutines instead of ~50, none pinning pool slots).
S8 — speed-test routes through pool.Manager + a singleflight-dedupe coordinator with a per-provider client cache and janitor sweep. Endpoint can no longer be rate-abused into exhausting a provider's connection budget.
S9 — MetadataVirtualFile.Close fires UsenetReader.Interrupt before contending for mvf.mu, so client disconnects mid-read return in microseconds instead of pinning the goroutine + connection slot for the full segment latency. Big one for Plex/Jellyfin scrubbing.
Test infrastructure

Deterministic fakepool seam (per-message latency / error injection, in-flight counters, gate primitive), segments payload generator, goroutines snapshot helper. Every fix is pinned by a regression test asserting MaxInFlight / BodyCalls / goroutine count directly.

Running result

Combined with [#595](https://github.com/javi11/altmount/issues/595) in my dev branch — solid through 24h+ playback + import workload on the gluetun/k8s setup that originally surfaced the conntrack storms. Go heap steady at ~26MB, no kernel WireGuard instability.

Clean merge against current main.

### DIFF
--- a/internal/api/provider_speedtest_handler.go
+++ b/internal/api/provider_speedtest_handler.go
@@ -2,8 +2,6 @@ package api
 
 import (
 	"context"
-	"crypto/tls"
-	"fmt"
 	"log/slog"
 	"time"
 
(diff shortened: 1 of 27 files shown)


javi11:
After merge the first PR this becomed corrupted, can you fix it @iPromKnight ?

iPromKnight:
> After merge the first PR this becomed corrupted, can you fix it @iPromKnight ?

yep - rebased on main :)

PR #602, opened 2026-05-20
fix(health): preserve SABnzbd category dirs in library sync

iPromKnight:
## Summary
- `library_sync.removeEmptyDirectories` walked `ImportDir` and deleted every empty subdirectory it found, guarded only against `MountPath`, `LibraryDir`, `/` and `.`. SABnzbd category folders (movies, tv, …) were not protected, so the moment a category went transiently empty between imports the sync wiped it.
- Radarr v6 has dropped the `Disabled Health Check Notifications` setting entirely (verified by source search — no occurrence of `disabledHealthChecks` in modern Radarr), so the resulting `RemotePathMappingCheck` error ("places downloads in `/mnt/.../movies` but this directory does not appear to exist") cannot be silenced client-side anymore. The only fix is for altmount to stop deleting the dirs.
- Added `buildProtectedImportDirs(cfg)` which derives the protected set from config: `ImportDir` itself, the optional `CompleteDir` level, every category dir (using `Dir||Name` fallback), the default category when no categories are configured, plus parents-of-nested-category-dirs (e.g. the `media` ancestor when a category `Dir` is `media/movies`). Wired into the existing `removeEmptyDirectories` guard with one extra membership check.

2026-05-20 at 21 37 06

2026-05-20 at 21 37 39

## Scope
Single source of truth for the deletion (`removeEmptyDirectories`) is `internal/health/library_sync.go` — verified by auditing every other `os.Remove` / `RemoveAll` site in the codebase:

- All other `os.Remove` calls operate on individual files (NZBs, symlinks) at the leaf, not directories.
- The arrs-webhook deletion path (`internal/api/arrs_handlers.go`) goes through `metadataService.DeleteDirectory`, which targets the metadata root tree (`ms.rootPath`), not `ImportDir`.

## Behavior preserved
Genuine empty release subfolders *inside* a category (e.g. `<ImportDir>/movies/Some.Release.2024/`) still get cleaned up. Only the configured category-level paths (and their structural ancestors) are protected.

## Test plan
- [x] Six failing-then-passing subtests covering every config shape that appears in the wild:
  - `complete_dir: ""` + explicit categories (by Name)
  - `complete_dir: ""` + category `Dir` overrides `Name`
  - `complete_dir: "complete"` + categories
  - `complete_dir: ""` + no categories (Default category dir)
  - `complete_dir: ""` + nested category `Dir` (`media/movies`)
  - `complete_dir: "/"` collapses to `ImportDir` (real-world config from this issue)
- [x] Three subtests for the `buildProtectedImportDirs` helper degrade cases: nil `ImportDir`, empty `ImportDir`, nil `cfg` — all return an empty set.
- [x] Each preservation test also asserts that empty release subfolders *under* the category still get cleaned up — no over-protection.
- [x] Full `go test -race ./internal/health/...` passes.
- [x] `golangci-lint ./internal/health/...` clean.
- [x] `go build ./...` succeeds.
- [x] Manual: verify Radarr v6 health check clears after deploying the patched build.

### DIFF
--- a/internal/health/library_sync.go
+++ b/internal/health/library_sync.go
@@ -1526,6 +1526,61 @@ func (lsw *LibrarySyncWorker) getLibraryPath(metaPath string, filesInUse map[str
 	return nil
 }
 
+// buildProtectedImportDirs returns the set of clean absolute paths under
+// ImportDir that must never be deleted by removeEmptyDirectories. This
+// covers ImportDir itself, the optional CompleteDir level, and every
+// configured SABnzbd category dir (plus the default category when no
+// categories are configured). Without this protection, transiently-empty
+// category folders get pruned between imports and arrs (Radarr v6+) raise
+// a permanent RemotePathMappingCheck health error.
+func buildProtectedImportDirs(cfg *config.Config) map[string]struct{} {
+	protected := map[string]struct{}{}
+	if cfg == nil || cfg.Import.ImportDir == nil || *cfg.Import.ImportDir == "" {
+		return protected
+	}
+
+	importDir := filepath.Clean(*cfg.Import.ImportDir)
+	protected[importDir] = struct{}{}
+
+	base := importDir
+	if cfg.SABnzbd.CompleteDir != "" {
+		base = filepath.Join(importDir, cfg.SABnzbd.CompleteDir)
+		protected[filepath.Clean(base)] = struct{}{}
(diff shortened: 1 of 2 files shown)


iPromKnight:
21 hours uptime. All fine and health check errors gone, after 30 imports each arr

PR #605, opened 2026-05-21
fix(sabnzbd): mark history slot Failed when reported path is missing

### LINKED ISSUE #596: SAB-history reports paths that do not exist on disk → ARR queues fill with FileNotFoundException
## Summary

altmount's SAB-history endpoint returns `status=Completed` for items
whose reported path doesn't exist on disk, causing Sonarr/Radarr to
import-loop with `FileNotFoundException` until manually unstuck.

The endpoint already detects this condition and logs it:

```
level=WARN msg="sabnzbd history: reported path does not exist on disk"
  item_id=17311
  storage_path=/complete/shows-DUTCH/Doug.Unplugs.S01E01.DUTCH.2160p.WEB.H265-REMBRANDT-xpost
  reported_path=/mnt/debrid/altmount-sym/complete/shows-DUTCH/Doug.Unplugs.S01E01.DUTCH.2160p.WEB.H265-REMBRANDT-xpost
```

But it still returns the item with the missing path and `status=Completed`. The ARRs enqueue the item and fail import.

## Environment

- altmount image: `ghcr.io/javi11/altmount:pinned-20260513`
- Import strategy: `SYMLINK`, `Import.ImportDir = /mnt/debrid/altmount-sym`
- FUSE mount: `/mnt/altmount` (rclone)
- Sonarr/Radarr use altmount as the SAB-compatible download client

## What's actually happening (verified)

Earlier I framed this as "altmount serves release B when release A was grabbed". That was wrong, there is no rename or content drift.

Verified across 10 currently-WARN'd items:

| Check | Result |
|---|---|
| FUSE file exists at `MountPath + storage_path` | yes, 10/10 |
| Symlink exists at `ImportDir + storage_path`   | no, 10/10  |
| FUSE serves a different release than storage_path advertises | no, 0/10 |

All 10 items have FUSE mtime between May 17 and May 19 03:57, before a permission issue on `ImportDir` was fixed locally. After the fix, about 5,955 symlinks were created on the day of the fix and ~1,300 per day since, so new imports create symlinks fine. These 10 are legacy items where symlink creation failed at import time and was never retried.

altmount's queue table keeps them with `status=Completed`, and `calculateHistoryStoragePath` returns the (non-existent) symlink path on every SAB-history poll. There is no self-healing path: the symlink never gets created, the item is never marked failed, and the ARRs see "Completed, path P" forever.

## Expected

When `calculateHistoryStoragePath` detects the path is missing:

1. (minimal) Return `status=Failed` for that slot in the SAB-history response, with a fail message pointing at the missing path. The ARR removes it cleanly through its normal failed-download path.
2. (better) Attempt to recreate the symlink first, then fall through to (1) if creation still fails.

(1) alone is enough to stop the loop. Happy to send a PR for (1).

## Repro

1. Configure altmount with `Import.ImportStrategy=SYMLINK` and an `Import.ImportDir` separate from `MountPath`.
2. Cause symlink creation to fail at import time (e.g. wrong perms on `ImportDir`).
3. Watch the affected items stay at `status=Completed` with `reported_path` on `ImportDir` even after perms are fixed. Each SAB-history poll fires the WARN; each ARR-side import fires `FileNotFoundException`.

## Workaround used locally

10-minute cron: `find /mnt/debrid/altmount-sym -xtype l -delete` on the symlink farm, plus per-ARR queue cleanup that blocklists the release and removes it. Doesn't fix the underlying altmount queue entry, just keeps it from saturating the ARR queues.

## Related

- #538 (closed 2026-04-26, "Fixed in last dev commit"): same WARN, the `*.mkv_<id>` filename-suffix variant. Adjacent, different fix surface.
- #586: queue removal triggering DownloadedMovieImportService. Different path.

firestaerter3:
## Summary

If altmount's `calculateHistoryStoragePath` finds that the reported symlink path doesn't exist on disk, the SAB-history response still ships the slot with `status=Completed`. Sonarr/Radarr then queue the item and hit `FileNotFoundException` on import.

This is the loop described in #596. The fix marks the slot as `Failed` when the path is missing, so the ARR removes it cleanly through its normal failed-download path.

## What changed

- `internal/api/sabnzbd_handlers.go`: `calculateHistoryStoragePath` now returns `(path string, exists bool)`. The two callers (`respondSABnzbdHistory`, `respondSABnzbdHistoryByIDs`) override the slot's `Status` to `Failed` and set a `FailMessage` when `!exists`.
- `internal/api/sabnzbd_types.go`: small `markHistorySlotMissing` helper to overwrite an existing slot's status.
- `internal/api/sabnzbd_types_test.go`: unit tests for the missing-path override, including pre-existing `fail_message` preservation and nil safety.

## What this does NOT do

It does not retry symlink creation. If `Import.ImportDir` is writable, new imports create symlinks normally; this only ensures historical items where creation already failed get reported to the ARR cleanly instead of looping forever.

## Test plan

- Tests added in `sabnzbd_types_test.go`.
- I couldn't run `go test` locally (Go 1.26.0 toolchain plus a DNS issue on my end fetching modules). Relying on CI here. Happy to iterate if anything fails.

Fixes #596.


firestaerter3:
Good catch, thanks. Fixed in 1932a1e.

The fix now gates `markHistorySlotMissing` on a per-item set (`liveCompleted`) populated only when the item came from the live queue with `Status=Completed`. Rows reconstructed from persistent `ImportHistory` via `importHistoryToQueueItem` are left untouched, so a past success whose file was later removed by ARR cleanup keeps its Completed status.

Applied to both `respondSABnzbdHistory` and `respondSABnzbdHistoryByIDs`.

javi11:
Dou you tested this on your setup and is working as expected? as first glance seems fine

javi11:
@firestaerter3 do this pr is still relevant?

firestaerter3:
Ah yes... working in my setup (running myself with this patch)

PR #608, opened 2026-05-22
fix(windows): remove remaining SYMLINK guards missed by #607

fatbob01:
## Summary

#607 removed the Windows SYMLINK guards from `symlink_creator.go` and `health_handlers.go` but two more remained that prevent altmount from starting or functioning on Windows:

- **`internal/config/manager.go`**: startup config validator rejects `import_strategy: SYMLINK` on Windows with `"SYMLINK is not supported on Windows; use STRM instead"` — the binary exits immediately without serving
- **`internal/health/library_sync.go`**: guard blocks symlink updates during library sync on Windows

Both blocks follow the same `runtime.GOOS == "windows"` pattern as the ones removed in #607. Also removed the now-unused `"runtime"` imports from both files.

## Test plan

- [x] `go build ./...` clean on Windows
- [x] altmount starts successfully with `import_strategy: SYMLINK` on Windows
- [x] Symlinks created correctly on import

### DIFF
--- a/internal/config/manager.go
+++ b/internal/config/manager.go
@@ -7,7 +7,6 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
-	"runtime"
 	"slices"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


fatbob01:
Found these while testing #607 on Windows — the binary wouldn't start at all due to the validator in manager.go rejecting SYMLINK at boot. The library_sync.go one would have surfaced later during library sync. Both follow the same pattern as what was removed in #607.

PR #613, opened 2026-05-24
fix(rar): fail fast on incomplete archives instead of falsely completing

evulhotdog:
Three bugs caused altmount to falsely report NZBs as successfully imported when RAR volumes were missing from the Usenet server:

1. ValidateSegmentIntegrity checked against PackedSize (sum of found parts) instead of Size (uncompressed full file size from RAR header). When a volume is missing, PackedSize shrinks and validation incorrectly passes.

2. patchMissingSegment created overlapping fake segments with wrong offsets, falsely inflating perceived coverage without adding real data.

3. Validation failures in aggregator.go were silently skipped by returning nil instead of propagating the error.

Changes:
- common.go: validate against content.Size, fallback to PackedSize only if Size is unavailable
- processor.go: fix patchMissingSegment to use correct StartOffset and SegmentSize
- aggregator.go: return errors from both validation paths instead of nil

Result: incomplete archives now fail during import with a clear error, allowing ARR apps to retry instead of trying to read a non-existent file.

### DIFF
--- a/internal/importer/archive/common.go
+++ b/internal/importer/archive/common.go
@@ -112,15 +112,18 @@ func ValidateSegmentIntegrity(ctx context.Context, content Content) error {
 			}
 		}
 	} else {
-		// For standard files, validate total segment coverage against PackedSize (if available)
+		// For standard files, validate total segment coverage against Size (uncompressed).
+		// Size comes from the first RAR volume header and is the correct full file size.
+		// PackedSize is the sum of actual parts found and can be truncated when volumes
+		// are missing, so we must validate against Size to detect incomplete archives.
 		var totalCovered int64
 		for _, seg := range content.Segments {
 			totalCovered += (seg.EndOffset - seg.StartOffset + 1)
 		}
 
-		expectedSize := content.PackedSize
+		expectedSize := content.Size
 		if expectedSize <= 0 {
-			expectedSize = content.Size
+			expectedSize = content.PackedSize
 		}
 
 		if expectedSize > 0 {
--- a/internal/importer/archive/rar/aggregator.go
+++ b/internal/importer/archive/rar/aggregator.go
@@ -252,6 +252,23 @@ func ProcessArchive(ctx context.Context, opts ProcessArchiveOptions) error {
 		}
 	}
 
+	// Compute total archive size and a critical-file threshold.
+	// Validation failures on files below the threshold are logged and skipped
+	// (e.g. .nfo, .srt, .jpg sidecars in a missing last volume) so that the
+	// import does not fail when only a tiny sidecar is incomplete.
+	// Validation failures on files at or above the threshold fail the archive.
+	var totalArchiveSize int64
+	for _, c := range rarContents {
+		if !c.IsDirectory {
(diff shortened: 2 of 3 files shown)


javi11:
but with this if the failing parts are in a file inside the rar that is not the video file it will make all the file fails

evulhotdog:
@javi11 what do you think is a better path to address it? I do see the conversation in the issue I created. Maybe we can wait to see what happens there.

evulhotdog:
@javi11 I don't like the hardcoded extensions, but I'm not sure what other way to tackle this...

```
RAR archive validation flow

 rarContents
      │
      ▼
 hasAllowedFiles? ─── no ──▶ FAIL (ErrNoAllowedFiles)
      │
     yes
      │
      ▼
 Build filesToProcess list
      │
      │  ┌─────────────────────────────────────┐
      │  │  Per file:                          │
      │  │    IsDirectory?       ─▶▶ skip      │
      │  │    IsAllowedFile?     ─▶▶ skip      │
      │  │      (ext check + sample filter)    │
      │  │    keep ▶ add to validation pool    │
      │  └─────────────────────────────────────┘
      │
      ▼
 Parallel validation goroutines
      │
      ├── isPreExtracted? ─▶▶ skip validation
      │
      ├── validateSegmentIntegrity
      │       (checks Size, not PackedSize)
      │       │
      │       ├── passes ─▶▶ ValidateSegmentsForFile
      │       │                  │
      │       │                  ├── passes ─▶▶ ✓ write metadata
      │       │                  │
      │       │                  └── fails ───┐
      │       │                               │
      │       └── fails ──────────────────────┤
      │                                       │
      │              ◀────────────────────────┘
      │              │
      │              ▼
      │         isSidecarFile?
      │      (.srt .sub .jpg .nfo etc.)
      │              │
      │       ┌──────┴──────┐
      │       │             │
      │      yes           no
      │       │             │
      │       ▼             ▼
      │   ⚠ warn, skip   ✗ FAIL archive
      │   return nil      return err
      │       │             │
      │       └─────┬───────┘
      │             │
      │             ▼
      │        filesProcessed > 0?
      │             │
      │      ┌──────┴──────┐
      │      │             │
      │     yes           no
      │      │             │
      │      ▼             ▼
      │   ✓ SUCCESS    FAIL (ErrNoFilesProcessed)

 KEY:
   ▶▶  = file skipped silently
   ✓   = success path
   ✗   = failure path
   ⚠   = sidecar skip (new behavior)
```

javi11:
So, if I understand well, what you will want is something like the allowed file extensions but that mark the file as failed if no video file is found available no?

evulhotdog:
@javi11 yes, but still being validated through the normal mechanisms, so we don't fail on a sample, nfo, etc. but do fail on the main file. 

I don't know if there's a better path to go here. To salvage what we can. 

It also may be a moot workaround, if we can implement par2 repairs on the fly.

javi11:
`It also may be a moot workaround, if we can implement par2 repairs on the fly.` this is not technically possible.

Ok what I can do is, a setting to fail if the imported dosn't have any video file that is not a sample

PR #622, opened 2026-05-27
fix(arrs): notify *arrs on import failure so they search again sooner

evulhotdog:
## Summary
When an import fails (e.g. RAR decode error, missing segments), AltMount previously let the failed item sit in the queue with `status: failed` without any notification to Sonarr/Radarr. The arrs would only discover the failure on their next periodic poll of the SABnzbd history API, which could be minutes later.

Now `HandleFailure` in the postprocessor coordinator triggers `RefreshMonitoredDownloads` on the relevant ARR type (determined from the item's category), the same way the success path already does via `broadcastToARRType`. This causes Sonarr/Radarr to check the download client immediately, see the failed status, and search for a replacement on their next search cycle.
- `broadcastToARRType` call added to `Coordinator.HandleFailure`, gated by the existing `SkipArrNotification` flag
- Logs a warning if the ARR type cannot be determined from the category, but does not block the existing fallback/cleanup flow

### DIFF
--- a/internal/importer/postprocessor/coordinator.go
+++ b/internal/importer/postprocessor/coordinator.go
@@ -156,6 +156,26 @@ func (c *Coordinator) HandleSuccess(ctx context.Context, item *database.ImportQu
 func (c *Coordinator) HandleFailure(ctx context.Context, item *database.ImportQueueItem, _ error) error {
 	cfg := c.configGetter()
 
+	// Notify ARR applications so they check for the failed download when they
+	// next query the SABnzbd history, rather than waiting for their periodic
+	// poll to discover it.
+	if !shouldSkipARRNotification(item) {
+		c.mu.RLock()
+		arrsService := c.arrsService
+		c.mu.RUnlock()
+
+		if arrsService != nil {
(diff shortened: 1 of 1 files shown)


internal/importer/postprocessor/coordinator.go
```diff
@@ -156,6 +156,26 @@ func (c *Coordinator) HandleSuccess(ctx context.Context, item *database.ImportQu
 func (c *Coordinator) HandleFailure(ctx context.Context, item *database.ImportQueueItem, _ error) error {
 	cfg := c.configGetter()
 
+	// Notify ARR applications so they check for the failed download when they
+	// next query the SABnzbd history, rather than waiting for their periodic
+	// poll to discover it.
+	if !shouldSkipARRNotification(item) {
```
javi11, line 169: This should be done if there is no fallback to sabnzbd; otherwise, the fallback will be ignored

evulhotdog: Ok, bumped it up so sabnzbd external call happens first!

javi11 (approved):

PR #631, opened 2026-05-30
fix(importer): add debug logging and progress updates during segment …

evulhotdog:
## Problem

NZB imports with large segment counts (e.g., 12,996 segments) get stuck at **"Validating segments 30%"** for hours without any visible progress or error. This happens because:

1. **No debug logging in the validation path** — the system silently checks segments with zero observability into what's happening
2. **Progress only updated on success** — the `progressTracker.Update()` call was gated behind `err == nil`, so when segments fail or timeout, the UI stays frozen at the starting percentage (30%)
3. **100% segment sampling on imports** — with `segment_sample_percentage: 100`, every single segment gets `STAT`-ed via NNTP. At 10 concurrent connections and 30s timeout per segment, a 13k-segment file can take 30+ minutes even when segments are missing
4. **No per-segment failure logging** — failures were swallowed without logging the segment ID or error, making it impossible to diagnose missing articles

### Symptoms

- Queue items seemingly stuck in `processing` status
- UI shows exactly "30%" for the entire duration
- Logs show only `Processing file` then nothing until completion/failure
- DB shows `processing` with no error message

## Root Cause

The `ValidateSegmentList` function in `internal/usenet/validation.go` used an `int32` counter named `validatedCount` that was **only incremented when `usenetPool.Stat()` returned successfully**. The `progressTracker.Update()` was called using this success-only counter. When segments are missing (returning `nntp: 430 no such article`) or timing out, the counter never advanced and the UI never moved.

Additionally, the entire validation pipeline had very little debug logging, making it impossible to distinguish between "hung" and "very slow" from the logs alone.

### 1. Track completion, not just success (`internal/usenet/validation.go`)

- Renamed `validatedCount` to `completedCount` and `successCount`
- `completedCount` increments after every `Stat()` call regardless of outcome
- `progressTracker.Update()` now uses `completedCount`, so the UI advances on every segment check

### 2. Added structured debug logging across the validation path

- **`internal/usenet/validation.go`** — `ValidateSegmentList` start/finish, per-segment success/failure with segment IDs
- **`internal/importer/validation/segments.go`** — `ValidateSegmentsForFile` start/finish with segment counts
- **`internal/importer/singlefile/processor.go`** — entry/exit of single-file validation
- **`internal/importer/multifile/processor.go`** — entry/exit of multi-file validation

### 3. Consistent logging conventions

Used `slog.DebugContext` and `slog.WarnContext` consistently, matching the codebase's structured logging conventions. No log values are interpolated into message strings.

## Files Changed

- `internal/usenet/validation.go`
- `internal/importer/validation/segments.go`
- `internal/importer/singlefile/processor.go`
- `internal/importer/multifile/processor.go`

## Verification

```bash
go build ./internal/importer/validation/ ./internal/usenet/ ./internal/importer/singlefile/ ./internal/importer/multifile/
go test ./internal/usenet/
Both pass cleanly. The binary was also deployed to a running container and verified that debug logs now appear in the logs.


javi11:
This debug logs will be way too spam, this should't happends, can you dm me the nzb and your verification config?

PR #623, opened 2026-05-27
feat(health): overhaul provider health dashboard with interactive cyber-HUD

drondeseries:
This PR introduces a comprehensive overhaul of the provider health dashboard:

1. **Interactive Cyber-HUD**: Overhauls the provider health page with a gorgeous visual dashboard including a provider topology map, interactive latency/speed charts, and custom provider-specific stats.
2. **Indexer Auditing**: Introduces detailed database schemas to monitor indexer import stats, queries, and success/failure ratios over time, with SQL migrations for both SQLite and Postgres.
3. **Automated *arr Webhook Registration**: Enables webhooks by default and automatically registers *arr instances upon detection to streamline post-processing without requiring manual user configuration.
4. **Enhanced Confirm Modal**: Improves confirm modals by adding support for custom text-verification inputs (e.g. typing a keyword/name to confirm destructive or critical actions).

PR #635, opened 2026-05-31
feat(health): restore indexer tracking from webhook Grab and OnImport events

drondeseries:
This pull request restores indexer tracking from webhook Grab and OnImport events. It logs persistent indexer failure when a file fails health checks and aligns the indexer stats query with the upstream schema. (Stacked on top of #623 dashboard overhaul).

### DIFF
--- a/internal/api/arrs_handlers.go
+++ b/internal/api/arrs_handlers.go
@@ -60,9 +60,9 @@ type ArrsWebhookRequest struct {
 		Id int64 `json:"id"`
 	} `json:"bookFile"`
 
-	EventType string `json:"eventType"`
+	EventType    string `json:"eventType"`
 	InstanceName string `json:"instanceName,omitempty"`
-	FilePath  string `json:"filePath,omitempty"`
+	FilePath     string `json:"filePath,omitempty"`
 	// For upgrades/renames, the file path might be in other fields or need to be inferred
 	Movie struct {
 		Id         int64  `json:"id"`
@@ -85,6 +85,11 @@ type ArrsWebhookRequest struct {
 		Path      string `json:"path"`
 	} `json:"episodeFile"`
 	DeletedFiles ArrsDeletedFiles `json:"deletedFiles,omitempty"`
+	DownloadId   string           `json:"downloadId,omitempty"`
+	Release      *struct {
+		Indexer      string `json:"indexer,omitempty"`
+		ReleaseTitle string `json:"releaseTitle,omitempty"`
+	} `json:"release,omitempty"`
 }
 
 func (req ArrsWebhookRequest) ToMetadata() model.WebhookMetadata {
@@ -244,6 +249,22 @@ func (s *Server) handleArrsWebhook(c *fiber.Ctx) error {
 	case "Test":
 		slog.InfoContext(c.Context(), "Received ARR test webhook")
 		return c.Status(200).JSON(fiber.Map{"success": true, "message": "Test successful"})
+	case "Grab":
+		if req.DownloadId != "" && req.Release != nil && req.Release.Indexer != "" {
+			indexerName := req.Release.Indexer
+			releaseTitle := req.Release.ReleaseTitle
+			s.importerService.StoreGrabbedIndexer(req.DownloadId, releaseTitle, indexerName)
+			slog.InfoContext(c.Context(), "Logged grabbed indexer from webhook", "download_id", req.DownloadId, "release_title", releaseTitle, "indexer", indexerName)
+			// Proactively update any existing queue item with this download ID
+			if err := s.queueRepo.UpdateQueueItemIndexerByDownloadID(c.Context(), req.DownloadId, indexerName); err != nil {
+				slog.WarnContext(c.Context(), "Failed to update indexer for existing queue item", "download_id", req.DownloadId, "indexer", indexerName, "error", err)
+			}
+			// In case the import already completed or failed (e.g. race condition), update history
+			if err := s.queueRepo.UpdateImportHistoryIndexerByDownloadID(c.Context(), req.DownloadId, indexerName); err != nil {
+				slog.DebugContext(c.Context(), "Failed to update indexer for import history (expected if not yet complete)", "download_id", req.DownloadId, "indexer", indexerName, "error", err)
+			}
+		}
+		return c.Status(200).JSON(fiber.Map{"success": true, "message": "Grab logged successfully"})
 	case "Download", "AlbumImport", "BookImport": // OnImport
 		isScanEvent = true
 		if req.EpisodeFile.Path != "" {
@@ -252,6 +273,22 @@ func (s *Server) handleArrsWebhook(c *fiber.Ctx) error {
 			pathsToScan = append(pathsToScan, req.MovieFile.Path)
 		} else if req.FilePath != "" {
 			pathsToScan = append(pathsToScan, req.FilePath)
+		}
+
+		// Update indexer name in database tables if present in webhook
+		if req.DownloadId != "" && req.Release != nil && req.Release.Indexer != "" {
+			indexerName := req.Release.Indexer
+			slog.InfoContext(c.Context(), "Logged indexer from OnImport webhook", "download_id", req.DownloadId, "indexer", indexerName)
+
+			// 1. Update queue if it still exists
+			if err := s.queueRepo.UpdateQueueItemIndexerByDownloadID(c.Context(), req.DownloadId, indexerName); err != nil {
+				slog.DebugContext(c.Context(), "Failed to update indexer for queue item", "download_id", req.DownloadId, "indexer", indexerName, "error", err)
+			}
+
+			// 2. Update import history if it has already completed
+			if err := s.queueRepo.UpdateImportHistoryIndexerByDownloadID(c.Context(), req.DownloadId, indexerName); err != nil {
+				slog.DebugContext(c.Context(), "Failed to update indexer for import history", "download_id", req.DownloadId, "indexer", indexerName, "error", err)
+			}
 		}
 	case "Rename":
 		isScanEvent = true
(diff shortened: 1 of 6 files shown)


internal/database/health_repository.go
```diff
@@ -1869,3 +1869,9 @@ func (r *HealthRepository) UpdateFileMetadata(ctx context.Context, id int64, met
 	_, err := r.db.ExecContext(ctx, query, metadata, id)
 	return err
 }
+
+// LogIndexerImport records a success or failure for an indexer persistently.
+func (r *HealthRepository) LogIndexerImport(ctx context.Context, indexer string, status string, errMsg string, downloadID *string) error {
+	return logIndexerImport(ctx, r.db, indexer, status, errMsg, downloadID)
```
javi11: 🔴 **HIGH — Misleading API: `downloadID` parameter is silently discarded**

The INSERT in `logIndexerImport` no longer includes `download_id`, but this public wrapper still accepts `downloadID *string` that is never forwarded. Every caller (including both `handleProcessingSuccess` and `handleProcessingFailure` in `service.go`) passes a value believing it will be stored.

**Fix**: Remove the parameter from this signature and all call sites, or restore the column in the INSERT. As-is, the parameter communicates a contract that no longer holds and will mislead future readers.

internal/importer/service.go
```diff
@@ -1504,3 +1549,63 @@ func (s *Service) RegenerateMetadata(ctx context.Context, mountRelativePath stri
...
+}
+
+// GetGrabbedIndexer retrieves a grabbed indexer by download ID or release title
+// It only returns a match if it was stored within the last 15 seconds.
+func (s *Service) GetGrabbedIndexer(downloadID string, releaseTitle string) (string, bool) {
+	now := time.Now()
+
+	check := func(key string) (string, bool) {
+		if val, ok := s.grabbedIndexers.Load(key); ok {
+			if info, isInfo := val.(grabbedIndexerInfo); isInfo {
```
javi11, line 1587: 🔴 **HIGH — 15-second TTL makes this cache ineffective in production**

The Grab webhook fires when the download client accepts the item. The Download/OnImport webhook — which triggers `AddToQueue` and ultimately `handleProcessingSuccess`/`handleProcessingFailure` — fires after the download *completes*, typically **minutes to hours** later. This 15-second window will almost always be expired by then.

The OnImport webhook handler already writes the indexer directly to the DB (`UpdateQueueItemIndexerByDownloadID`, `UpdateImportHistoryIndexerByDownloadID`), so the in-memory path is only needed as a race-condition guard (Grab and Import arrive in the same second). If that's the intent, document it. But a 15-second TTL is too short even for that — a fast download can take more than 15 s to process.

**Options:**
1. Remove the TTL entirely — rely on `downloadID` for dedup (it's stable and unique); clear entries only after successful lookup in `handleProcessingSuccess`/`handleProcessingFailure`.
2. Increase TTL to something realistic (e.g. 24 h) and add a background sweeper to prevent the map from growing unbounded.

internal/database/indexer_repository.go
```diff
@@ -100,9 +100,8 @@ func deleteIndexerStats(ctx context.Context, db DBQuerier, indexer string) (int6
 }
 
 func updateIndexerStatsByDownloadID(ctx context.Context, db DBQuerier, downloadID string, indexer string) error {
-	query := `UPDATE indexer_import_stats SET indexer = ? WHERE download_id = ? AND indexer = 'Unknown'`
-	_, err := db.ExecContext(ctx, query, indexer, downloadID)
-	return err
+	// No-op: we keep upstream schemas clean and already associate download_ids in queue/history tables
```
javi11, line 103: 🟡 **MEDIUM — No-op function still called from three sites**

`updateIndexerStatsByDownloadID` now does nothing but return nil. It is still called from both the Grab and OnImport blocks in `arrs_handlers.go`, and likely from the queue repository as well. Dead call sites create confusion about intent and waste a function-call round-trip.

**Fix**: Remove all call sites. If the no-op is temporary while the schema settles, leave a TODO with an issue link so it doesn't become permanent dead code.

internal/api/arrs_handlers.go
```diff
@@ -244,6 +249,21 @@ func (s *Server) handleArrsWebhook(c *fiber.Ctx) error {
...
+			indexerName := req.Release.Indexer
+			releaseTitle := req.Release.ReleaseTitle
+			s.importerService.StoreGrabbedIndexer(req.DownloadId, releaseTitle, indexerName)
+			slog.InfoContext(c.Context(), "Logged grabbed indexer from webhook", "download_id", req.DownloadId, "release_title", releaseTitle, "indexer", indexerName)
+			// Proactively update any existing queue item with this download ID
+			if err := s.queueRepo.UpdateQueueItemIndexerByDownloadID(c.Context(), req.DownloadId, indexerName); err != nil {
+				slog.WarnContext(c.Context(), "Failed to update indexer for existing queue item", "download_id", req.DownloadId, "indexer", indexerName, "error", err)
+			}
+			// In case the import already completed or failed (e.g. race condition), update history and stats
+			_ = s.queueRepo.UpdateImportHistoryIndexerByDownloadID(c.Context(), req.DownloadId, indexerName)
```
javi11: 🟡 **MEDIUM — Silent error vs logged error: inconsistent within same handler**

A few lines above, `UpdateQueueItemIndexerByDownloadID` logs a `WarnContext` on failure. Here the same class of errors (`UpdateImportHistoryIndexerByDownloadID` and `UpdateIndexerStatsByDownloadID`) are silently discarded with `_ =`. The race-condition comment explains *why* these may fail, but observability still matters — a `slog.DebugContext` at minimum would let you spot unexpected failure modes without noise.

internal/importer/service.go
```diff
@@ -1504,3 +1549,63 @@ func (s *Service) RegenerateMetadata(ctx context.Context, mountRelativePath stri
...
+
+	if downloadID != "" {
+		s.grabbedIndexers.Store(downloadID, info)
+	}
+	// Sanitize and use release title as an alternative fallback key
+	if releaseTitle != "" {
+		cleanTitle := strings.TrimSpace(strings.ToLower(releaseTitle))
+		cleanTitle = strings.TrimSuffix(cleanTitle, ".gz")
+		cleanTitle = strings.TrimSuffix(cleanTitle, ".nzb")
+		s.grabbedIndexers.Store(cleanTitle, info)
```
javi11, line 1576: 🟡 **MEDIUM — `grabbedIndexers` sync.Map leaks entries that are never looked up**

`StoreGrabbedIndexer` writes two keys per call: one for `downloadID` and one for `cleanTitle`. Expired entries are only pruned lazily inside `check()` when that exact key is looked up. If a Grab webhook arrives but the download is never imported (cancelled, replaced, etc.), both keys accumulate forever — a slow memory leak in a long-running service with many queued downloads.

Also, when `check(downloadID)` succeeds and deletes the `downloadID` key, the sibling `cleanTitle` key is left behind, and vice versa.

**Fix**: Add a periodic sweeper (e.g. `time.AfterFunc` at store time, or a background goroutine on startup) that calls `Range` and deletes expired entries. Or use a single canonical key (`downloadID`) and drop the title-based fallback — the OnImport webhook provides the `downloadId` anyway.

internal/importer/service.go
```diff
@@ -1022,6 +1032,22 @@ func (s *Service) resolveCategoryPath(category string) string {
 
 // handleProcessingSuccess handles all steps after successful NZB processing
 func (s *Service) handleProcessingSuccess(ctx context.Context, item *database.ImportQueueItem, resultingPath string) error {
+	// Log persistent indexer statistic
```
javi11, line 1038: 🟡 **MEDIUM — Magic string `"Unknown"` should be a named constant**

`"Unknown"` is used as a sentinel indexer name in at least three files: `worker.go`, `service.go` (here and in `handleProcessingFailure`), and likely the DB layer. A typo divergence would silently cause stats to double-count under two different keys.

```go
const indexerUnknown = "Unknown"
```

Define it once (in the `database` package where the column lives, or a shared `constants` package) and reference it everywhere.

internal/health/worker.go
```diff
@@ -474,6 +474,16 @@ func (hw *HealthWorker) prepareUpdateForResult(ctx context.Context, fh *database
 			update.Status = database.HealthStatusCorrupted
 			sideEffect = func() error {
 				slog.ErrorContext(ctx, "File permanently marked as corrupted after repair retries exhausted", "file_path", fh.FilePath)
+				
+				// Log failure against the indexer if known
```
javi11, line 478: 🟢 **LOW — Trailing whitespace on blank lines**

Both added `sideEffect` blocks contain blank lines with leading tabs before and after the new `if fh.Indexer != nil` block. `gofmt` will flag these. Run `gofmt -w internal/health/worker.go` to clean up.

javi11 (commented):
## Code Review — feat(health): restore indexer tracking from webhook events

| Severity | Count | Blocking |
|----------|-------|---------|
| 🔴 HIGH | 2 | Yes |
| 🟡 MEDIUM | 4 | Recommend fix |
| 🟢 LOW | 1 | No |

**Blocking items** (see inline comments for details):
1. **Misleading `downloadID` parameter** in `LogIndexerImport` / `logIndexerImport` — accepted but silently dropped after the schema change. All callers believe it is stored.
2. **15-second TTL** on `grabbedIndexers` — the Grab→Import gap is measured in minutes to hours in real use; this window expires before it can ever be useful via the cache path.

The overall approach (in-memory grab→indexer map with DB fallbacks on OnImport) is sound. The OnImport DB-write path is a good belt-and-suspenders fallback. The two HIGH items are correctness/observability issues that would silently produce wrong or missing indexer stats in production.

drondeseries:
Hi @javi11,

I have addressed 100% of your excellent review feedback, fully rebased the branch on top of the newly merged upstream  (which now has the dashboard merge), and force-pushed. 

The PR is now perfectly clean with **exactly 1 conflict-free commit**!

### 🛠️ Summary of Changes:
1. **Misleading `downloadID`**: Removed the parameter from `LogIndexerImport` and its internal helper `logIndexerImport` across `indexer_repository.go` and `health_repository.go`, and updated all call sites (`worker.go` and `service.go`).
2. **15-second TTL**: Increased the lookup window TTL in `GetGrabbedIndexer` to a safe **24 hours** to handle queue/processing lag.
3. **Memory Leak in `grabbedIndexers`**: Added a periodic background goroutine sweeper `runGrabbedIndexerSweeper` running on a 1-hour ticker to prune expired mappings.
4. **Sentinel String**: Defined a global `database.IndexerUnknown = "Unknown"` constant in `models.go` and refactored all references.
5. **No-op Calls**: Deleted all dead round-trip call sites to `UpdateIndexerStatsByDownloadID`.
6. **Observability**: Replaced the silent `_ =` ignores in the webhook handler blocks with structured `slog.DebugContext` calls to surface failure modes without adding noise.
7. **Whitespace**: Cleaned up the trailing whitespace/leading tabs in `worker.go` and ran `gofmt -w`.

Thank you for the thorough and incredibly sharp review!

PR #642, opened 2026-05-31
feat(fuse): add no_mod_time option to suppress unstable virtual filesystem mtimes

DrMxrcy:
## Problem

Virtual filesystems (NFS, FUSE-backed cloud mounts, segment-cache layers) can report inconsistent or rapidly-changing modification times for files whose content has not actually changed. This happens because: 

- Segment caches update file metadata as segments are fetched, changing the stored mtime
- Directory nodes have no natural mtime and previously returned `time.Now()` on every `Getattr` call
- Some underlying storage backends don't preserve mtimes accurately across mounts/remounts

Media servers and indexing tools that rely on mtime to detect file changes will incorrectly treat these files as modified on every scan, triggering unnecessary re-indexing, metadata re-extraction, and in some cases entering re-processing loops that stall the entire scan.

## Solution

Adds a `no_mod_time: true` option to `FuseConfig`. When enabled, the FUSE layer reports Unix epoch (zero) for all `mtime`, `ctime`, and `atime` values instead of the underlying virtual filesystem value.

This gives any consumer of the mount a stable, consistent timestamp. Since the files themselves are read-only (NZB archives on Usenet don't change), reporting epoch is semantically accurate — the content never changes, so the mtime should never change either.

### Changes

- **`internal/config/manager.go`**: adds `NoModTime bool` (`no_mod_time` in YAML/JSON) to `FuseConfig`
- **`internal/fuse/backend/hanwen/utils.go`**: `fillAttr` accepts `noModTime bool`; returns zero timestamps when set
- **`internal/fuse/backend/hanwen/dir.go`**: `Dir` carries `noModTime`, propagated to all child `Dir` and `File` nodes created during `Lookup` and `Mkdir`
- **`internal/fuse/backend/hanwen/file.go`**: `File` carries `noModTime`, applied in `Getattr`
- **`internal/fuse/backend/hanwen/backend.go`**: passes `FuseConfig.NoModTime` to `NewDir` at mount time
- **`frontend/src/types/config.ts`**: adds `no_mod_time` to `FuseConfig` interface
- **`frontend/src/components/config/MountConfigSection.tsx`**: adds "Suppress modification times" checkbox to the FUSE Compatibility section

### Configuration

```yaml
fuse:
  no_mod_time: true
```

Defaults to `false` — no behaviour change for existing configurations.

### Notes

- The flag is named to mirror the equivalent option in rclone (`--no-modtime`) for familiarity
- Only the hanwen (Linux) backend is affected; the cgofuse backend has its own attr path
- On first mount after enabling, consumers that already stored a non-zero mtime will see one mtime transition (stored value → epoch); after that all stats return epoch and no further changes are detected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/frontend/src/components/config/MountConfigSection.tsx
+++ b/frontend/src/components/config/MountConfigSection.tsx
@@ -978,6 +978,7 @@ function FuseMountSubSection({ config, isRunning, onFormDataChange }: FuseSubSec
 		max_read_ahead_mb: 128,
 		async_buffer_size_mb: 16,
 		async_buffer_max_total_mb: 256,
+		no_mod_time: false,
 	});
 
 	useEffect(() => {
@@ -1139,6 +1140,20 @@ function FuseMountSubSection({ config, isRunning, onFormDataChange }: FuseSubSec
 							/>
 							<span className="label-text text-xs">Allow other users to access mount</span>
 						</label>
+					</fieldset>
+					<fieldset className="fieldset">
+						<legend className="fieldset-legend">Compatibility</legend>
+						<label className="label cursor-pointer justify-start gap-3">
+							<input
+								type="checkbox"
+								className="checkbox checkbox-primary checkbox-sm"
+								checked={formData.no_mod_time ?? false}
+								onChange={(e) => updateField({ no_mod_time: e.target.checked })}
+								disabled={isRunning}
+							/>
+							<span className="label-text text-xs">Suppress modification times (report epoch for all files)</span>
+						</label>
+						<p className="label">Enable when media servers misinterpret virtual filesystem mtimes as file changes</p>
 					</fieldset>
 				</div>
 			</div>
--- a/frontend/src/components/config/MountConfigSection.tsx
+++ b/frontend/src/components/config/MountConfigSection.tsx
@@ -1151,9 +1151,8 @@ function FuseMountSubSection({ config, isRunning, onFormDataChange }: FuseSubSec
 								onChange={(e) => updateField({ no_mod_time: e.target.checked })}
 								disabled={isRunning}
 							/>
-							<span className="label-text text-xs">Suppress modification times (report epoch for all files)</span>
+							<span className="label-text text-xs">Suppress modification times</span>
 						</label>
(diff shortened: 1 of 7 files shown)


javi11:
**LOW — `atime` zeroed alongside `mtime`/`ctime`**

In `utils.go`, all three timestamps are set to zero when `noModTime` is true. The option name `no_mod_time` implies only modification-time suppression, but access time (`atime`) is also zeroed — a slightly broader side-effect.

Pre-existing code already conflated `atime = mtime`, so this isn't a regression. And for a read-only NZB mount, `atime` tracking isn't useful. The PR description does mention all three values explicitly, so this is intentional — worth surfacing so users who rely on `find -atime` or similar tools on the mount know what to expect.

Suggestion: add a brief note to the `NoModTime` config comment:

```go
// NoModTime reports epoch for all timestamps (mtime, ctime, atime); prevents media servers re-scanning on unstable VFS mtimes.
NoModTime bool `yaml:"no_mod_time" mapstructure:"no_mod_time" json:"no_mod_time"`
```

Are you sure we want this?

DrMxrcy:
> **LOW — `atime` zeroed alongside `mtime`/`ctime`**
> 
> In `utils.go`, all three timestamps are set to zero when `noModTime` is true. The option name `no_mod_time` implies only modification-time suppression, but access time (`atime`) is also zeroed — a slightly broader side-effect.
> 
> Pre-existing code already conflated `atime = mtime`, so this isn't a regression. And for a read-only NZB mount, `atime` tracking isn't useful. The PR description does mention all three values explicitly, so this is intentional — worth surfacing so users who rely on `find -atime` or similar tools on the mount know what to expect.
> 
> Suggestion: add a brief note to the `NoModTime` config comment:
> 
> ```go
> // NoModTime reports epoch for all timestamps (mtime, ctime, atime); prevents media servers re-scanning on unstable VFS mtimes.
> NoModTime bool `yaml:"no_mod_time" mapstructure:"no_mod_time" json:"no_mod_time"`
> ```
> 
> Are you sure we want this?

Good catch on the comment — updated it to explicitly call out all three timestamps and added a note about find -atime behavior so users aren't surprised.

PR #643, opened 2026-06-01
Cleanup codebase, optimize hot paths, and resolve Sportarr indexer

yoshitaka420:


yoshitaka420:
## What this PR does

Two independent areas (bundled here, easy to split if you'd prefer separate PRs):

1. **Fix: Sportarr imports no longer logged as `Unknown` in Indexer Health**
2. **Codebase cleanup + hot-path optimizations** (net **−1,600 lines**)

---

### 1. Sportarr indexer attribution

**Problem.** Every Sportarr download showed up in Indexer Health as `Unknown`.

**Why.** Indexer attribution for the *arrs comes from the Grab/Import webhook AltMount auto-registers via `golift.io/starr`. Sportarr isn't starr-compatible, so that webhook can't be auto-registered — and checking Sportarr's source confirmed it's a dead end anyway: Sportarr only ever fires one notification (`OnDownload`, at import) and its payload carries **no indexer and no downloadId**. The indexer is only exposed on Sportarr's native queue (`/api/queue`).

**Fix (three parts):**
- **Decode tolerance** (the actual blocker) — current Sportarr serializes `status` as a numeric enum, `statusMessages` as a `[]string`, and `downloadClient` as an object, but the queue struct expected string / object-array / string. Any one mismatch failed the whole `json.Unmarshal`, so the queue read errored before anything else ran. Added tolerant unmarshalling that accepts both current and legacy (Servarr) shapes, with table tests for each.
- **Import-time resolution** (deterministic) — the indexer-health stat is written once at import completion and never rewritten, so a periodic poll can lose the race. `arrs.Service.ResolveIndexerByDownloadID` now looks the indexer up from Sportarr's queue by download ID synchronously, as the final fallback in the importer's success/failure paths.
- **Periodic capture** (best-effort) — the cleanup worker also stamps the indexer onto the queue/history rows as it polls.

**Caveats:** rows already logged `Unknown` aren't backfilled (the stat table has no download_id to re-key). Sportarr queue *cleanup* is separately degraded on current builds (its queue projection drops `outputPath`/`trackedDownloadState`) — out of scope here.

---

### 2. Cleanup + optimizations

- **Dead code removed (~700 lines, all verified unused):** `encryption/nonce.go`, `usenet/utils.go`/`errors.go`, `auth/api_key_middleware.go`, `rclonecli/stats.go`, plus dead methods across slogutil, health/indexer repos, fuse, rclone, httpclient, config.
- **Unified API responses / security:** migrated ~90 inline `fiber.Map` handler responses to the `response.go` builders. Besides consistency, this closes a production detail leak — those handlers returned raw `err.Error()` as `details`, which the builders suppress via `safeDetails()` outside dev.
- **DB:** `AddBatchToQueue` prepares its statement once instead of per row; the Postgres dialect rewrite is memoized; the `file_health` point-lookups share one column list + scan helper.
- **Frontend:** wrapped HealthPage/QueuePage list handlers in `useCallback` to restore `React.memo` on rows; deduplicated `ProviderChart`/`ProviderSpeedChart` into a shared component (~300 lines); consolidated `formatFileSize`→`formatBytes`; typed `api/client` error handling (removed `as any`); misc standards fixes.
- Minor: `go mod tidy`, a `tagged switch` lint fix.

---

### Verification
`make` (proto + `go mod tidy` + golangci-lint + `go test -race ./...`) passes for all changed packages; frontend `tsc` is clean. The new Sportarr decode tests prove `GetQueue` extracts `downloadId` + `indexer` from both the current and legacy queue shapes.

internal/arrs/clients/sportarr.go
```diff
@@ -33,19 +34,120 @@ type SportarrStatusMessage struct {
...
+	} else {
+		*s = sportarrStatus(fmt.Sprintf("%d", n))
+	}
+	return nil
+}
+
+// sportarrStatusMessages accepts either the Servarr object shape
+// [{title,messages}] (older builds) or a bare string array (current Sportarr),
+// so the queue decode doesn't fail outright when the shape changes.
+type sportarrStatusMessages []SportarrStatusMessage
```
javi11, line 107: is this really needed? can we support just the new one and make the people update to the latest?

yoshitaka420: I'll leave that up to you... if you want only newer one then I can modify it accordingly.

internal/nzbfilesystem/metadata_remote_file.go
```diff
@@ -864,6 +869,7 @@ type interruptSlot struct{ i readerInterrupter }
 // Callers must hold mvf.mu. Pass nil to clear.
 func (mvf *MetadataVirtualFile) setReader(r io.ReadCloser) {
```
javi11, line 870: nice one, i will refactor this next, since i don't like this approach

PR #645, opened 2026-06-01
Present a real SABnzbd identity on outbound NZB fetches (incl. across Prowlarr redirects)

yoshitaka420:
## Background

AltMount emulates a SABnzbd download client. A real SABnzbd's NZB grabber (urllib `_build_request`) sends a very specific, minimal set of headers when fetching an NZB:

- `User-Agent: SABnzbd/<version>`
- `Accept-Encoding: gzip`
- URL-embedded basic auth (userinfo), **never** an `X-Api-Key` header

Some indexers reject grabs that don't look like a genuine SABnzbd client. AltMount's outbound fetches didn't fully match this, so grabs could be refused — especially through Prowlarr, which most indexers require to issue a **302 redirect** to the indexer's own download URL.

## Changes

**1. SABnzbd User-Agent on every outbound NZB fetch**
- `addurl` handler: replaced the generic `altmount` UA with `SABnzbd/<version>`, routed the fetch through the proxy-aware external client, and redacted query strings (which carry indexer API keys) in logs.
- Prowlarr `DownloadNZB`: sends `SABnzbd/<version>` alongside the API key.
- `nzblnk` already defaulted to `SABnzbd/<version>`; updated frontend help text to describe the new default and override semantics.

**2. Header parity across Prowlarr's 302 redirect**
- Added a `CheckRedirect` policy that strips `X-Api-Key` once a redirect leaves the original Prowlarr host, so the indexer-facing request carries exactly the headers a real SABnzbd sends — and the Prowlarr key never leaks to a third party.
- The key is retained for same-host hops; Go's 10-redirect limit is preserved.
- Go already adds `Accept-Encoding: gzip` (transport compression) and derives basic auth from URL userinfo, covering the remaining SABnzbd headers.

**3. Weekly version refresh**
- `versionRefreshTTL` changed from 24h → 7 days, so the latest SABnzbd version is fetched at most once per week.

## Testing
- Wire-level tests covering all three outbound paths (`addurl`, Prowlarr `DownloadNZB`, `nzblnk`) plus URL/API-key redaction.
- Redirect tests proving the `SABnzbd/<version>` UA survives a Prowlarr 302 download redirect, and that `X-Api-Key` is dropped on cross-host redirects, for both the Prowlarr client and the `addurl` handler.

### DIFF
--- a/frontend/src/components/config/NzblnkConfigSection.tsx
+++ b/frontend/src/components/config/NzblnkConfigSection.tsx
@@ -61,12 +61,14 @@ export function NzblnkConfigSection({
 						className="input input-bordered w-full min-w-0 max-w-full bg-base-100 font-mono text-sm"
 						value={formData.user_agent ?? ""}
 						readOnly={isReadOnly}
-						placeholder="Mozilla/5.0 ... (leave empty for default)"
+						placeholder="SABnzbd/<version> (leave empty for default)"
 						onChange={(e) => handleInputChange("user_agent", e.target.value)}
 					/>
 					<p className="label min-w-0 max-w-full whitespace-normal break-words text-base-content/70 text-xs">
 						HTTP User-Agent sent when searching and downloading from public NZB indexers (e.g.
-						nzbking.com, nzbindex.com). Leave empty to use the built-in default.
+						nzbking.com, nzbindex.com). Leave empty to identify as a current SABnzbd release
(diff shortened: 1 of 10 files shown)


yoshitaka420:
Everything works fine from my testing but I want you to confirm and test this functionality from your end as well before merging it to main since this is a high risk change.

PR #667, opened 2026-06-08
fix(importer): namespace failed NZB path to avoid UNIQUE constraint collision

dclipca:
### Problem

`MoveToFailedFolder` moves a failed NZB to `<failedDir>/<basename>` and then runs `UPDATE import_queue SET nzb_path = ?`. Because `import_queue.nzb_path` is `UNIQUE`, two failed items that share a basename (e.g. a re-acquired release reusing the same NZB filename) collide on the second move:

```
failed to update DB with new NZB path: failed to update queue item nzb path: UNIQUE constraint failed: import_queue.nzb_path
```

The item still reaches `Failed`, but the failed-folder bookkeeping errors out and the on-disk NZB and the DB row drift apart.

### Fix

Namespace the failed copy with the queue item ID (`<id>-<basename>`) when the plain destination is already taken, keeping the original name otherwise. Extracted into a small `uniqueFailedNzbPath` helper with a unit test covering repeated basename collisions.

### Testing

`go test ./internal/importer/` and `go vet ./internal/importer/` pass; gofmt-clean.

### DIFF
--- a/internal/importer/failed_path_test.go
+++ b/internal/importer/failed_path_test.go
@@ -1,10 +1,14 @@
 package importer
 
 import (
+	"context"
+	"log/slog"
 	"os"
 	"path/filepath"
 	"testing"
 
+	"github.com/javi11/altmount/internal/config"
+	"github.com/javi11/altmount/internal/database"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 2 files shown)


dclipca:
@javi11

javi11:
could you add a test when you click retry in a failed item with this change? also try it manually. I have the suspect that it will break that functionality

dclipca:
Added the retry coverage in c693329, and I dug into whether namespacing can actually break retry — short version: it can't, because of the existing early-return in `MoveToFailedFolder`.

When you retry a failed item and it fails again, its `NzbPath` is already inside `failedDir`, so the guard short-circuits before the namespacing path is ever used:

```go
newPath := uniqueFailedNzbPath(failedDir, item.NzbPath, item.ID) // stat only, not yet used

// Avoid moving if already in failed folder (e.g. retry of failed item)
if filepath.Dir(item.NzbPath) == failedDir {
    return nil // retry-refail bails here, before the rename + DB update
}
```

The subtle bit (and I think what you were sensing) is that `uniqueFailedNzbPath` *would* namespace the item's own existing file to `<id>-<name>` if it ever reached the rename — but it never does on retry, because the guard returns first. So the namespaced `newPath` is computed and discarded.

Two tests drive the real `MoveToFailedFolder` (not just the helper) to lock this in:

- `TestMoveToFailedFolder_RetryOfFailedItemIsNoop` — item already in the failed folder: path unchanged, file intact, **no** `<id>-` duplicate created.
- `TestMoveToFailedFolder_RetryOfFailedItemInCategoryIsNoop` — same guarantee for a categorized item whose failed copy lives in `failedDir/<category>`.

I left `database` nil in the test service on purpose: the retry path must return before touching the repository, so if a future change ever makes it fall through to the DB update, the test nil-panics instead of silently regressing.

`gofmt -l` clean, `go vet ./internal/importer/` clean, full `go test ./internal/importer/` passes.

On the manual side: we run this fork in production (a streamer instance plus an importer instance). This UNIQUE-constraint collision isn't theoretical for us — we hit it on the importer when a re-acquired release reused an NZB basename and the failed-folder bookkeeping errored out, drifting the on-disk NZB from the DB row. That's what motivated the PR. (Full disclosure: our prod importer is still on the pre-fix commit, so I can't yet claim the patch itself has soaked in prod — happy to report back once it has.)

PR #669, opened 2026-06-09
fix(health): match Windows library paths in GetUnhealthyFiles

fatbob01:
## Problem

On Windows, altmount stores `library_path` values in SQLite with backslashes (e.g. `C:\rclone\show-torrents\Show\Season 1\Episode.mkv`). The `GetUnhealthyFiles` query only matched a forward-slash prefix:

```sql
OR (library_path IS NOT NULL AND library_path LIKE ?)
```

Where the bound parameter was `C:/rclone/%`. This never matches a backslash-stored path, so health checks never fire for Windows users — all 9000+ files stay permanently `pending`.

## Fix

Add a second `OR` clause binding the backslash variant:

```sql
OR (library_path IS NOT NULL AND (library_path LIKE ? OR library_path LIKE ?))
```

With parameters `C:/rclone/%` and `C:\rclone\%`.

Also simplifies the prefix-building logic using `strings.TrimRight` to handle both separator styles consistently.

## Impact

**No behavioral change on Linux/macOS** — backslash paths never appear there, so the new clause never matches. Windows users get health checks working correctly.

## Test

Added `TestGetUnhealthyFiles_MatchesWindowsLibraryPath` to `health_repository_test.go` — inserts a record with a Windows-style backslash `library_path` and asserts `GetUnhealthyFiles` returns it.

### DIFF
--- a/internal/database/health_repository.go
+++ b/internal/database/health_repository.go
@@ -210,7 +210,7 @@ func (r *HealthRepository) GetUnhealthyFiles(ctx context.Context, limit int, str
 		  AND status NOT IN ('repair_triggered', 'checking')
 		  AND (
 			  ? = 'NONE' 
-			  OR (library_path IS NOT NULL AND (library_path LIKE ? OR library_path LIKE ?))
+			  OR (library_path IS NOT NULL AND (library_path LIKE ? ESCAPE '!' OR library_path LIKE ? ESCAPE '!'))
 			  OR (last_error LIKE '%failed to unmarshal metadata%')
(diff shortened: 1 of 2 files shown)


fatbob01:
Closing to reopen with a more complete fix. While investigating Windows health check failures in production, we found a second related issue: records written on Windows before the path-separator fix were stored with backslashes in file_path. normalizeHealthPath only stripped the leading slash, so exact WHERE file_path = ? lookups silently miss these legacy records. Will reopen shortly with both fixes together.

PR #673, opened 2026-06-09
fix(health): Windows path handling in health checks

fatbob01:
## Problem

On Windows, health checks were silently skipping all files. Three related issues:

**1. `GetUnhealthyFiles` never matched Windows library paths**
The query filtered `library_path LIKE 'C:/rclone/%'` (forward slash) but on Windows, SQLite stores library_path values with backslashes (`C:\rclone\show\ep.mkv`). No records ever matched, leaving all files permanently `pending`.

**2. Backslash as escape character in PostgreSQL**
The alternate backslash LIKE pattern (`C:\rclone\%`) uses `\%` which PostgreSQL interprets as an escaped literal `%` rather than a wildcard, so the pattern would never match on Postgres installs.

**3. `normalizeHealthPath` didn't normalize backslash separators**
Records written on Windows before path-separator fixes were stored with backslashes in `file_path` (e.g. `tv\Show\ep.mkv`). `normalizeHealthPath` only stripped the leading slash, so exact `WHERE file_path = ?` lookups silently missed these legacy records across all 19+ callers.

## Fix

**1.** Add a second `OR` clause with a backslash-variant prefix so both formats match:
```sql
OR (library_path LIKE ? ESCAPE '!' OR library_path LIKE ? ESCAPE '!')
```

**2.** Use `ESCAPE '!'` on both clauses — `!` is the escape character so `\` has no special meaning and `\%` correctly matches a literal backslash followed by any suffix on both SQLite and PostgreSQL.

**3.** Add `strings.ReplaceAll(p, `\`, "/")` to `normalizeHealthPath` so backslash paths are normalized to forward slashes before any DB read or write. No-op on Linux/macOS.

## Impact

No behavioral change on Linux/Docker — backslash paths never appear there. Windows users get health checks working correctly for the first time.

## Tests

Added `TestGetUnhealthyFiles_MatchesWindowsLibraryPath` — inserts a record with a Windows-style backslash `library_path` and asserts `GetUnhealthyFiles` returns it correctly.

## Verified

Confirmed working in production on a Windows install with 9800+ files — all transitioned from permanently `pending` to `healthy` after this fix.

### DIFF
--- a/internal/database/health_repository.go
+++ b/internal/database/health_repository.go
@@ -30,7 +30,8 @@ func NewHealthRepository(db *sql.DB, d Dialect) *HealthRepository {
 // repair state machine relies on. Every method that writes or matches on file_path
 // funnels through here.
 func normalizeHealthPath(p string) string {
-	return strings.TrimPrefix(p, "/")
+	p = strings.TrimPrefix(p, "/")
+	return strings.ReplaceAll(p, `\`, "/")
 }
 
 // escapeLikePrefix escapes the LIKE metacharacters in a literal prefix so it can be
--- a/internal/database/health_repository.go
+++ b/internal/database/health_repository.go
@@ -237,7 +237,7 @@ func (r *HealthRepository) GetUnhealthyFiles(ctx context.Context, limit int, str
 		  AND status NOT IN ('repair_triggered', 'checking', 'corrupted')
 		  AND (
 			  ? = 'NONE' 
-			  OR (library_path IS NOT NULL AND (library_path LIKE ? OR library_path LIKE ?))
+			  OR (library_path IS NOT NULL AND (library_path LIKE ? ESCAPE '!' OR library_path LIKE ? ESCAPE '!'))
 			  OR (last_error LIKE '%failed to unmarshal metadata%')
 			  OR (last_error LIKE '%failed to read file metadata%')
 			  OR (last_error LIKE '%no ARR instance found%')
--- a/internal/database/health_repository.go
+++ b/internal/database/health_repository.go
@@ -237,7 +237,7 @@ func (r *HealthRepository) GetUnhealthyFiles(ctx context.Context, limit int, str
 		  AND status NOT IN ('repair_triggered', 'checking', 'corrupted')
 		  AND (
 			  ? = 'NONE' 
-			  OR (library_path IS NOT NULL AND library_path LIKE ?)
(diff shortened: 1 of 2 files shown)


fatbob01:
## Why this is needed (root cause) and why it's safe on Linux/Docker

**Root cause — native Windows only.** For the SYMLINK strategy, `library_path` is populated by the library sync from a filesystem walk of `library_dir`: `result.Symlinks[cleanTarget] = path` in `internal/health/library_sync.go`, where `path` comes straight from `filepath.WalkDir` and is intentionally *not* run through `filepath.ToSlash` (the adjacent STRM branch is, but symlink targets are kept verbatim because repairs must use the exact path the *arr expects). On native Windows that walk returns backslash paths, so `library_path` is stored as e.g. `C:\library\Show\S01E01.mkv`.

`GetUnhealthyFiles` then gates checks on `library_path LIKE <library_dir>/%`. The old prefix was built as `libraryDir + "/%"` — a forward-slash-only pattern. On Windows that becomes `C:\library/%`, which can never match the backslash-stored value, so every imported file stays `pending` forever and no health check ever fires. Docker/Linux installs are unaffected because their paths are forward-slash natively.

**Why the fix is safe on Linux/Docker:**

- **Dual LIKE pattern** — the forward-slash pattern (`/library/%`) is unchanged and does all the matching on Linux. The added backslash pattern (`\library\%`) can never match a forward-slash path, so it is an inert no-op there.
- **`ESCAPE '!'`** — required so the backslash pattern works on PostgreSQL, where `\` is the default LIKE escape character (so a trailing `\%` would otherwise mean a *literal* `%` and never match). It is a no-op on SQLite and irrelevant to forward-slash paths, so it *prevents* a Postgres regression rather than introducing risk.
- **`normalizeHealthPath` backslash→slash** — canonicalizes `file_path` on read/write so a leading-slash or backslash variant of the same virtual path cannot split into two rows and defeat the repair-retry budget. Backslashes do not appear in real Linux media paths, so this is effectively Windows-only in practice.

**Net effect:** no behavioral change on Linux/macOS/Docker, fixes a hard stall on native Windows, and hardens Postgres. The change is layout-independent — it matches both separator styles regardless of whether `import_dir` and `library_dir` share a root. Confirmed against three independent native-Windows installs that were all stuck `pending` on stock and recovered with this change.

PR #695, opened 2026-06-19
fix(importer): NZB subjects wrapped in quotes break multi-volume RAR enumeration

evulhotdog:
## Problem

Some NZBs wrap the `<file subject>` filename in surrounding quotes:

```xml
<file subject="'Some.Show.S03.WEBRip.EAC3.2.0.1080p.x265-GRP.part001.rar'">
```

The external `nzbparser` library extracts the filename verbatim, keeping the surrounding quotes. AltMount then stored `'Some.Show.S03...part001.rar'` (with the leading and trailing `'`) as the filename for every volume in the set.

The trailing `'` after `.rar` defeats every RAR volume-detection regex in the codebase, all of which anchor on `$` (end-of-string), so `.rar'` never matches `.rar`. As a result:

- `IsRarFile()` returned `false` for all continuation volumes (parts 002–N), so they were not propagated as RAR archives.
- `SetKey()` returned `("", false)` for every volume, so `GroupArchivesByBaseName` fell back to using the full quoted filename as the grouping key, producing **one group per volume** instead of one group for the whole set.
- `rardecode` received N single-part "archives" and errored `bad volume number` on every non-first part.
- Only `part001.rar` was partially read, and its single inner file was renamed to the NZB release name and emitted as a "healthy" virtual file, even though it was backed by only a fraction of the release's segments.

A secondary bug amplified the damage: when volume following stopped early, `convertAggregatedFilesToRarContent` produced a `Content` whose `Segments` covered far fewer bytes than the file's declared `Size` from the RAR header, yet still emitted it as healthy. Downstream tools (Sonarr/Radarr) then marked the episode as downloaded and stopped searching for a valid copy.

### 1. Strip surrounding quotes at the `nzbparser` boundary (anti-corruption layer)

Per review feedback, quote-stripping moved out of the scattered leaf call sites (`fileinfo.selectBestFilename`, the `rar.SetKey` wrapper) into a **single normalization point at the `nzbparser.Parse` boundary**, so every consumer sees a canonical filename:

- **`internal/importer/utils/nzbtrim/nzb_trim.go`** — `TrimSurroundingQuotes(s)` pure stdlib leaf. Strips one matched pair of `'...'`/`"..."` after trimming whitespace; idempotent; leaves embedded apostrophes (`It's.A.Wonderful.Life.part01.rar`) intact.
- **`internal/importer/parser/parser.go`** — `SanitizeNzbFilenames(n)` mutates `NzbFile.Filename` in place immediately after `nzbparser.Parse`, before any consumer reads it. The raw `Subject` is preserved untouched.
- Wired at all three `nzbparser.Parse` call sites: `parser.ParseFile` (`parser.go:83`), the pre-parse fast-fail probe (`processor.go:405`), and `service.go:1575`.
- The priority-0 `subjectHeader` candidate (`parser.go:247`) is also trimmed, since it is derived from the same poster-controlled `Subject`.
- **Removed** the duplicate `trimSurroundingQuotes` from `archive/rar/utils.go` (`SetKey` is now a pure passthrough) and `fileinfo.go` (`selectBestFilename` is pure priority selection). `rarname.SetKey` / `rar.SetKey` now carry a documented precondition: filenames must arrive canonical, no cleaning performed.
- **Legacy read-time shim** in `filesystem/volume_index.go:volumeKey()` trims quotes on read, so metadata persisted before this fix (which may still carry quotes) resolves correctly without a migration.

This puts normalization at the ingestion seam — the boundary between untrusted poster-controlled input and AltMount's internal model — covering all consumers (pre-parse fast-fail probe, parser, persisted metadata, serve-time volume following, user-visible filename) with one pass, and keeping the `rarname`/`rar` leaves pure interpretation.

### 2. Omit partially-covered inner files instead of emitting them as healthy

In `convertAggregatedFilesToRarContent` (`archive/rar/processor.go`), after segment mapping, the mapped bytes are summed and compared to the file's declared `Size`. When `Size - mappedBytes > maxSingleSegmentSize` (~800 KB, the typical Usenet segment size), the file is omitted with a `WarnContext` log entry rather than emitted as healthy.

`maxSingleSegmentSize` was promoted from a buried inline constant inside `patchMissingSegment` to a documented package-level const, reused by both the new coverage check and the existing single-segment patch logic.

This is the only import-time guard for truncated volume following: the downstream `ValidateSegmentIntegrity` and `ValidateSegmentsForFile` checks validate against `content.PackedSize`, and rardecode's `TotalPackedSize` accumulates only over the parts actually followed (`archive_info.go:134`), so it shrinks in lockstep with the segments on truncation and both pass. Only `TotalUnpackedSize` (`archive_info.go:93`, set once from the first block header = full declared size) stays at the full size, which is what this check anchors to.

## Tests

All changes were TDD — tests written RED first, then made GREEN.

**`TestTrimSurroundingQuotes`** (`nzbtrim/nzb_trim_quotes_test.go`) — single/double quotes, nested pairs, embedded apostrophe guard, idempotency.

**`TestSanitizeNzbFilenames`** (`parser/sanitize_test.go`) — asserts `NzbFile.Filename` is cleaned in place while the raw `Subject` is preserved.

**`TestExtractRarBaseName`** / **`TestGroupArchivesByBaseNamePaddedWidth`** (`rar/utils_test.go`) — repurposed from the old quote-specific tests; now assert recognized volumes resolve to their base key and that parts collapse into one group regardless of zero-padding width (inputs are canonical; quotes are stripped upstream). The embedded-apostrophe guard is retained.

**`TestConvertAggregatedFilesToRarContentOmitsTruncated`** (`rar/processor_test.go`) — three synthetic inner files where `file1.bin` declares 2 MB but has only 200 bytes of mapped segments; asserts `file1.bin` is omitted and the other two are emitted.

## Scope

- No changes to `sabnzbd_handlers.go` or any API/config/UI surface.
- `GroupArchivesByBaseName` algorithm body is unchanged; the fix is a pre-processing normalization at the parser boundary.
- `rarname.SetKey` / `rar.SetKey` are unchanged in logic — only their precondition is now documented, and the redundant cleaning wrapper was removed.
- The per-file coverage check is forward-only; previously-imported files are not affected.
- 12 files changed, +285/-4.


internal/importer/archive/rar/processor.go
```diff
@@ -509,6 +515,26 @@ func (rh *rarProcessor) convertAggregatedFilesToRarContent(ctx context.Context,
...
+		// archive), a file can claim its full size while only a fraction is mapped;
+		// emitting it surfaces an unplayable file that still reports "healthy", so
+		// *ARRs treat the file as complete and stop searching for a
+		// good copy. Drop any file short by more than one segment's worth
+		// (maxSingleSegmentSize) so it stays absent, wanted, and looking for a valid copy.
+		var mappedBytes int64
+		for _, s := range fileSegments {
+			mappedBytes += s.EndOffset - s.StartOffset + 1
+		}
+		if rc.Size > 0 && rc.Size-mappedBytes > maxSingleSegmentSize {
```
javi11: Is this really needed? I think is already verified later

javi11 (changes requested):

internal/importer/archive/rar/processor.go
evulhotdog: The later checks don't actually catch this scenario.

The downstream validators (I found both ValidateSegmentIntegrity and ValidateSegmentsForFile) compare segments against `PackedSize`, and not `Size`. And in rardecode, TotalPackedSize is accumulated only over the parts that were actually followed, so when volume following stops early, PackedSize actually shrinks in in tandem with the segments. Both checks compare partial-vs-partial and pass.

Only TotalUnpackedSize stays at the full declared size from the header, which is what this check uses. So it's the only import-time safety check for a partially-followed file.

checkVolumeCoverage doesn't cover it either because it's a set-level aggregate (80% of total bytes), so one short inner file in an otherwise-complete set sails through.

For example....

Lets say that there's a 2 MB file across 10 volumes, and you only 2 followed, so you get a PackedSize roughly 200 KB, segments roughly 200 KB, both downstream checks pass (200 KB vs 200 KB). This check: 200 KB vs 2 MB -> drops it.

The only overlap is if zero parts are followed, ValidateSegmentIntegrity falls back to Size and catches it, so this really catches the partial case.

Keep in mind, this validates file by file, and won't just dismiss all contents within the rar, only the ones it knows it can't use, and helps salvage what files _are_ accessible and available.

javi11: But, do you actually have a problem now? so you see failing imports because of this?

evulhotdog: I've removed the changes from this PR, so we talk about it at a later time.

I would like to get the quote changes merged!

javi11 (approved):

PR #708, opened 2026-06-21
feat(health): database conflict merging, worker safety, cache invalidation, and repair optimizations

drondeseries:
Implements key stability features for AltMount's health monitoring and file repair pipelines: Overhauls SQLite database UPSERT and relinking logic in health_repository.go to safely merge health records; solves scheduling races and deadlocks in the scheduled repair loop; invalidates cached ARR metadata when receiving import webhooks; and adds performance limits to SABnzbd queue queries.

### DIFF
--- a/internal/api/arrs_handlers.go
+++ b/internal/api/arrs_handlers.go
@@ -16,6 +16,16 @@ import (
 	"github.com/javi11/altmount/internal/database"
 )
 
+// ArrsInstanceRequest represents a request to create/update an arrs instance
+type ArrsInstanceRequest struct {
+	Name              string `json:"name"`
+	Type              string `json:"type"`
+	URL               string `json:"url"`
+	APIKey            string `json:"api_key"`
+	Category          string `json:"category"`
+	Enabled           bool   `json:"enabled"`
+	SyncIntervalHours int    `json:"sync_interval_hours"`
+}
 // ArrsWebhookRequest represents a webhook payload from Radarr/Sonarr
 type ArrsWebhookRequest struct {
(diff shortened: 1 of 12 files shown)


internal/health/library_sync.go
```diff
@@ -1062,6 +1066,25 @@ func (lsw *LibrarySyncWorker) getAllMetadataFiles(ctx context.Context) ([]string
 
 		// Only include .meta files
 		if !d.IsDir() && strings.HasSuffix(d.Name(), ".meta") {
+			nameWithoutMeta := strings.TrimSuffix(d.Name(), ".meta")
```
javi11: i think samples should be removed from import not here

internal/health/library_sync.go
```diff
@@ -1062,6 +1066,25 @@ func (lsw *LibrarySyncWorker) getAllMetadataFiles(ctx context.Context) ([]string
...
+			if sampleProofPattern.MatchString(nameWithoutMeta) {
+				// Convert to mount relative path to query lite cache
+				mountRelPath := lsw.metaPathToMountRelativePath(path)
```
javi11: this should be on import

internal/health/library_sync.go
drondeseries: Removed the redundant sample/proof filtering logic from library sync since it is already handled at the import boundary.

internal/health/library_sync.go
drondeseries: Removed in the latest commit.

javi11 (approved):

PR #712, opened 2026-06-23
Fix: deduplicate SABnzbd history by nzo_id in addition to name

jackster1209:
**Summary**

- The SABnzbd history endpoint deduplicates items by NZB name, but the same logical release can carry different name strings across the live-queue and persistent-history sources (e.g. <id>-Name.nzb.gz vs the clean NzbName)
- Name-only dedup let these through as two separate slots that share the same nzo_id, causing *arr apps to ingest the same download twice
- This built duplicate queue rows in *arr and triggered a .Single() collision on queue delete
- Adds a secondary seenNzoIDs guard that mirrors the nzo_id key logic already used in ToSABnzbdQueueSlot/ToSABnzbdHistorySlot, ensuring a release can't appear twice regardless of how its name is represented in each source

**Test plan**

1. Import a release via SABnzbd API and confirm it appears only once in the *arr queue
2. Confirm no duplicate nzo_id entries appear in the SABnzbd history response when the same item exists in both the live queue (as Completed) and persistent history
3. Verify *arr queue delete no longer throws a .Single() collision error for previously affected items

Co-authored with Claude Sonnet 4.6

### DIFF
--- a/internal/api/sabnzbd_handlers.go
+++ b/internal/api/sabnzbd_handlers.go
@@ -831,6 +831,20 @@ func (s *Server) handleSABnzbdHistory(c *fiber.Ctx) error {
 	// Combine and deduplicate by NZB Name
 	// Priority goes to items still in the queue (as they have more metadata)
 	seenNames := make(map[string]bool)
+	// Also dedupe by the nzo_id we will actually emit. The same logical release
+	// can carry two different name strings across the live-completed-queue and
+	// persistent-history sources (e.g. "<id>-Name.nzb.gz" vs the clean NzbName),
+	// so name-only dedup leaks it through as two slots sharing one nzo_id. *arr
+	// then ingests the same download_id twice, building duplicate queue rows and
+	// tripping a .Single() collision on queue delete. nzoKey mirrors the id logic
+	// in ToSABnzbd{Queue,History}Slot so the guard matches the emitted value.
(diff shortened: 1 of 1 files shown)


internal/api/sabnzbd_handlers.go
```diff
@@ -831,6 +831,20 @@ func (s *Server) handleSABnzbdHistory(c *fiber.Ctx) error {
...
 	seenNames := make(map[string]bool)
+	// Also dedupe by the nzo_id we will actually emit. The same logical release
+	// can carry two different name strings across the live-completed-queue and
+	// persistent-history sources (e.g. "<id>-Name.nzb.gz" vs the clean NzbName),
+	// so name-only dedup leaks it through as two slots sharing one nzo_id. *arr
+	// then ingests the same download_id twice, building duplicate queue rows and
+	// tripping a .Single() collision on queue delete. nzoKey mirrors the id logic
+	// in ToSABnzbd{Queue,History}Slot so the guard matches the emitted value.
+	seenNzoIDs := make(map[string]bool)
+	nzoKey := func(downloadID *string, id int64) string {
```
javi11, line 842: This can be an isolated method/util, no need to put it here

javi11:
can you modify that? if not i will need to close the pr

jackster1209:
I can work on modifying this later tonight when I have free time to do so, sorry for the delay.

PR #732, opened 2026-06-25
fix(arrs): blocklist-only season+episode fallback for renamed Sonarr …

EsmailELBoBDev2:
…episodes

A renamed-but-still-owned Sonarr episode (HasFile=true) whose episode-file id could not be matched by path dead-ended in ErrPathMatchFailed and was permanently condemned (the existing metadata.Episodes fallback only helps when webhook episode ids are present).

Add a season+episode fallback parsed from the path: resolve the owned episode(s) by their S/E and re-search them. This path is blocklist-only and delete-safe by construction:
- it only collects an episode id when ep.HasFile is true (owned episodes only);
- it calls the blocklist helper with fileID=0, and the helper now skips file-based history matching for a 0 sentinel so a missing file id can never blocklist the wrong release;
- it triggers only an EpisodeSearch. DeleteEpisodeFileContext is unreachable from this fallback — it lives solely inside the targetEpisodeFileID > 0 branch, which is never taken when this else branch runs.

### DIFF
--- a/internal/arrs/scanner/manager.go
+++ b/internal/arrs/scanner/manager.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log/slog"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


internal/arrs/scanner/manager.go
```diff
@@ -959,6 +983,38 @@ func (m *Manager) triggerSonarrRescanByPath(ctx context.Context, client *sonarr.
...
+		// otherwise dead-end in ErrPathMatchFailed and be permanently condemned. Resolve
+		// the owned episode(s) by the S/E parsed from the path and re-search them.
+		//
+		// CRITICAL DELETE SAFETY: this path is BLOCKLIST-ONLY. We pass fileID=0 to the
+		// blocklist helper — which then skips file-based history matching so a missing
+		// file id can't blocklist the wrong release — and fall through to trigger ONLY an
+		// EpisodeSearch below. DeleteEpisodeFileContext is unreachable from here: it lives
+		// solely inside the `targetEpisodeFileID > 0` branch above, which is never taken
+		// when this else branch runs (targetEpisodeFileID == 0).
+		if len(episodeIDs) == 0 {
```
javi11: i don´t really want this feature, in my case there is a lot of times where an episode can not be matched because of the releases in my languauge even i will say season packs and this will put that into sonarr blocklist

EsmailELBoBDev2: > i don´t really want this feature, in my case there is a lot of times where an episode can not be matched because of the releases in my languauge even i will say season packs and this will put that into sonarr blocklist

~~so close this PR? or do something diff or add a fallback?~~

Update: I asked claude to solve your issue so please see it and if you want to close PR no problem

EsmailELBoBDev2: > i don´t really want this feature, in my case there is a lot of times where an episode can not be matched because of the releases in my languauge even i will say season packs and this will put that into sonarr blocklist

Totally fair, that's exactly why I made it opt-in and off by default. It won't touch the setup unless someone explicitly enables it

So the season-pack and foreign-release case you're describing stays untouched out of the box.

For my use case the renamed-but-owned episodes were dead-ending as permanently corrupted, so I wanted an escape hatch. But I get that it's risky as a default. If you'd rather not carry an opt-in flag for it at all, I'm fine closing this one and keeping it on my fork.

PR #734, opened 2026-06-25
fix(arrs): use targeted Radarr movie lookup instead of full library scan

EsmailELBoBDev2:
The ID-based Radarr repair path fetched the entire movie library and looped to find one movie by internal id. On large libraries that full GetMovies call can exceed the HTTP client timeout, error out, and wrongly condemn the file.

Replace it with a targeted GetMovieByIDContext lookup. A failed/empty lookup (e.g. the movie was removed and re-added with a new internal id) falls through to the existing path-based matching instead of erroring. The smart-repair guard (a different healthy file -> ErrEpisodeAlreadySatisfied, no delete) is preserved.

### DIFF
--- a/internal/arrs/scanner/manager.go
+++ b/internal/arrs/scanner/manager.go
@@ -567,39 +567,41 @@ func (m *Manager) triggerRadarrRescanByPath(ctx context.Context, client *radarr.
 
 	// ID-Based Precision: If we have the movie ID from metadata, use it
 	if metadata != nil && metadata.Movie != nil && metadata.Movie.Id > 0 {
-		movies, err := m.data.GetMovies(ctx, client, instanceName)
-		if err != nil {
-			return fmt.Errorf("failed to get movies from Radarr for ID lookup: %w", err)
-		}
-		for _, movie := range movies {
-			if movie.ID == metadata.Movie.Id {
-				targetMovie = movie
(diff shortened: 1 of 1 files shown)


internal/arrs/scanner/manager.go
```diff
@@ -567,39 +567,41 @@ func (m *Manager) triggerRadarrRescanByPath(ctx context.Context, client *radarr.
...
-					slog.InfoContext(ctx, "ID-Based Precision: Using metadata movie ID for Radarr repair",
-						"movie_id", metadata.Movie.Id)
-					if movie.HasFile && movie.MovieFile != nil {
-						targetMovieFileID = movie.MovieFile.ID
+		// Targeted lookup by internal Radarr movie ID instead of fetching the ENTIRE
+		// movie library and scanning it. On large libraries the full GetMovies call can
+		// exceed the HTTP client timeout, error out, and wrongly condemn the file. A
+		// failed/empty targeted lookup (e.g. the movie was removed/re-added with a new
+		// internal id) falls through to the path-based matching below rather than erroring.
+		movie, lookupErr := client.GetMovieByIDContext(ctx, metadata.Movie.Id)
```
javi11, line 613: I understand your point, but the meta.movie.id is not always there you should fallback to the old method

EsmailELBoBDev2: done! please re-check

PR #735, opened 2026-06-25
fix(arrs): resolve re-added Radarr movies by stable TMDB id, delete-s…

EsmailELBoBDev2:
…afely

A removed-and-re-added Radarr movie gets a new internal id, so both the metadata-id lookup and the library path scan miss it and the corrupt file is wrongly condemned. Add a fallback that resolves the movie by its stable TMDB id via a targeted GetMovieContext(TMDBID) lookup.

Delete-safety is enforced through a single shared matcher, radarrFileMatchesTarget (exact path / basename / .strm-stripped / relative-suffix), which BOTH the library path scan and the TMDB-id fallback now route through so the rule cannot drift apart:
- file present AND matches the corrupt target -> blocklist + delete + search;
- file present but a DIFFERENT healthy file   -> ErrEpisodeAlreadySatisfied, no delete;
- no file yet                                 -> search only, no delete.

DeleteMovieFilesContext stays gated behind targetMovieFileID > 0, which the TMDB fallback only sets when the matcher confirms the target.

### DIFF
--- a/internal/arrs/scanner/manager.go
+++ b/internal/arrs/scanner/manager.go
@@ -553,6 +553,46 @@ func (m *Manager) TriggerDownloadScan(ctx context.Context, instanceType string) 
 	}
 }
 
+// radarrFileMatchesTarget reports whether the movie's current file is the corrupt
+// target identified by filePath/relativePath. It is the single source of truth for
+// the Radarr delete-safety decision: both the library path scan and the TMDB-id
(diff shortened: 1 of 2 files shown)


javi11:
add some test for this

EsmailELBoBDev2:
> add some test for this

done, please check

javi11 (approved):

PR #737, opened 2026-06-25
fix(health): tolerate transient streaming misses before condemning

EsmailELBoBDev2:
A single transient streaming miss (NNTP-430 / ArticleNotFound) immediately condemned a healthy file: the streaming error path marked the metadata corrupted, moved it to the safety folder and forced an immediate ARR repair.

Re-verify before condemning: on a streaming failure, schedule a high-priority health re-check (pending, scheduled now) instead of condemning, and leave the metadata in place so a healthy file is never hidden on one miss.

Also wire the existing acceptable_missing_segments_percentage config field (which previously had no backend reader) into the health checker: condemn only when the missing-segment ratio exceeds the threshold. The default of 0 preserves the historical strict behaviour.

### DIFF
--- a/internal/config/accessors.go
+++ b/internal/config/accessors.go
@@ -39,6 +39,21 @@ func (c *Config) GetSegmentSamplePercentage() int {
 	return c.Health.SegmentSamplePercentage
 }
 
+// GetAcceptableMissingSegmentsPercentage returns the percentage of sampled segments
+// that may be missing before a file is condemned as corrupted during a health check.
+// The default of 0 preserves the historical strict behaviour (any missing segment
+// condemns). The value is clamped to the [0, 100] range.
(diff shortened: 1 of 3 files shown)


javi11:
did you saw any case where you received a trancient 430? I never saw that, just couple of minutes after you just upload something to usenet and the article is not propagated

EsmailELBoBDev2:
> did you saw any case where you received a trancient 430? I never saw that, just couple of minutes after you just upload something to usenet and the article is not propagated

Yeah you're right, looked again and most of mine are just genuinely incomplete files, not transient 430s. My bad on that part.

The bit I'd keep is the config fix though: acceptable_missing_segments_percentage exists but nothing in the backend actually reads it, so the tolerance setting does nothing right now. Default 0 keeps current behavior. Happy to drop the re-verify half and keep just that if you want.

PR #757, opened 2026-07-02
perf(segcache): load catalog in background to avoid blocking startup

evulhotdog:
**Problem:** Startup hangs while `loadCatalog()` stats every cached segment on disk synchronously (one `os.Stat` per ~750KB segment, can be tens of thousands of calls).

**Fix:** Moved catalog loading from the `NewSegmentCache` constructor into `Manager.Start` as a background goroutine. Startup no longer blocks; the cache serves misses from Usenet until the catalog hydrates. Merge-under-lock prevents a `Put` during load from being clobbered.

### DIFF
--- a/internal/nzbfilesystem/segcache/cache.go
+++ b/internal/nzbfilesystem/segcache/cache.go
@@ -53,11 +53,16 @@ func NewSegmentCache(cfg Config, logger *slog.Logger) (*SegmentCache, error) {
 		return nil, fmt.Errorf("segcache: create cache dir %s: %w", cfg.CachePath, err)
 	}
 
-	return &SegmentCache{
+	c := &SegmentCache{
 		items:  make(map[string]*cacheEntry),
 		config: cfg,
 		logger: logger,
-	}, nil
+	}
+	// Gate Puts until LoadCatalog runs and clears this. Set at construction (not
+	// in Start's goroutine) so a segment downloaded before the async load runs is
+	// skipped rather than clobbered by the wholesale map assignment.
+	c.loading.Store(true)
+	return c, nil
 }
 
 // Has reports whether the segment is present in the cache (no disk I/O).
--- a/internal/nzbfilesystem/segcache/cache.go
+++ b/internal/nzbfilesystem/segcache/cache.go
@@ -42,6 +42,7 @@ type SegmentCache struct {
 	logger    *slog.Logger
 	totalSize int64
 	dirty     atomic.Bool
+	loading   atomic.Bool
 }
 
 // NewSegmentCache creates a new segment cache. It does NOT load any existing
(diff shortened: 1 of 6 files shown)


javi11:
@evulhotdog nice one, but don't you think that blocking the put while loading can cause cpu leak if there is a lot of put waiting to succed? maybe we should ignore the put while loading

evulhotdog:
@javi11 yeah, making it a no-op is probably easier to handle and reduces complexity. I was just trying to think about continuity and ensuring that we caching wherever we can. In general, altmount is a long running process, so this should be an edge case where you're streaming right away.

I was just tired of waiting 5 minutes to load my cache each time, since it stat's each file!

I simplified things, let me know what you think.

internal/nzbfilesystem/segcache/manager.go
```diff
@@ -86,9 +86,20 @@ func NewManager(cfg ManagerConfig, logger *slog.Logger) (*Manager, error) {
...
+// catalog load (run here, not in NewManager, so boot is not blocked by the
+// per-segment stat loop). Stop waits for all of them.
 func (m *Manager) Start(_ context.Context) {
-	m.wg.Add(2)
+	m.wg.Add(3)
+	// Gate Puts here on the caller's goroutine, before source.Swap wires the cache
+	// live, so a segment downloaded during cold load is skipped rather than
+	// clobbered by the wholesale map assignment. Load itself runs async so the
+	// per-segment stat loop does not block boot; Stop() waits via wg.
+	m.cache.loading.Store(true)
```
javi11: this should be inside the method LoadCatalog

evulhotdog: The NewSegmentCache now sets the var upon it being created, and then LoadCatalog's existing defer Store(false) clears it once completed.

evulhotdog:
@javi11 ready for ya again.

internal/nzbfilesystem/segcache/cache.go
```diff
@@ -55,8 +58,10 @@ func NewSegmentCache(cfg Config, logger *slog.Logger) (*SegmentCache, error) {
 		config: cfg,
 		logger: logger,
 	}
-	c.loadCatalog()
-
+	// Gate Puts until LoadCatalog runs and clears this. Set at construction (not
+	// in Start's goroutine) so a segment downloaded before the async load runs is
+	// skipped rather than clobbered by the wholesale map assignment.
+	c.loading.Store(true)
```
javi11: This should be in the LoadCatalog method

internal/nzbfilesystem/segcache/cache.go
```diff
@@ -245,8 +257,15 @@ func (c *SegmentCache) SaveCatalog() error {
 	return nil
 }
 
-func (c *SegmentCache) loadCatalog() {
+// LoadCatalog hydrates the in-memory catalog from catalog.json on disk, statting
+// each .seg file and dropping entries whose data is missing. Put is gated off
+// (see the loading flag) for the duration, so the load can assign the map
+// wholesale without racing a concurrent writer. Clears the gate when done.
+func (c *SegmentCache) LoadCatalog() {
+	defer c.loading.Store(false)
```
javi11, line 263: So here 
```suggestion
    c.loading.Store(true)
	defer c.loading.Store(false)
```

evulhotdog: Done!

internal/nzbfilesystem/segcache/cache.go
evulhotdog: Moved!

PR #793, opened 2026-07-23
fix(security): harden auth comparisons, CORS default, and remove dead salt code

WhispersOfJ:
## Summary
Found while doing an independent source-level security/resource-leak audit of this repo. This PR bundles the small, low-risk fixes from that pass; two higher-impact findings (an SSRF path via the SABnzbd ARR-credential auth flow, and `IsAdmin` not being enforced on any destructive route) are intentionally **not** in this PR — they change real request-handling behavior and deserve a design discussion, so I'll open them as separate issues instead of forcing a fix.

- **Constant-time credential comparisons**: `internal/api/parsers.go`'s API-key-override check and `internal/webdav/adapter.go`'s Basic Auth check both used plain `==`, which short-circuits on the first differing byte. Switched both to `crypto/subtle.ConstantTimeCompare`. The WebDAV path is the more realistic target — it's hit on every request from every media client, a real sample count to average network jitter over.
- **Exact-match auth-skip paths**: `RequireAuthWithSkip` (`internal/auth/middleware.go`) matched configured skip-paths with `strings.HasPrefix` instead of exact equality. All configured skip paths (`/auth/login`, `/auth/register`, etc.) are exact endpoints, so this was never intentional prefix matching — just a latent landmine where a future route like `/auth/login-audit` would silently bypass auth by string-prefix coincidence.
- **CORS default**: `internal/api/server.go` defaulted `AllowOrigins` to `"*"` while also setting `AllowCredentials: true` whenever neither `COOKIE_DOMAIN` nor `api.allowed_origins` was configured — a very plausible default/quickstart deployment. Wildcard-origin + credentialed-CORS is a combination the Fetch/CORS spec exists to forbid; depending on how the CORS middleware version in use handles it, this can mean any origin gets credentialed access to the API. Credentials are now only allowed once an explicit origin is configured.
- **Removed dead `DirectAuthSalt`**: generated (or read from `DIRECT_AUTH_SALT`) and stored on `auth.Config`, but never read anywhere else in the codebase. bcrypt already salts stored password hashes on its own, so this field only implied an extra security layer that didn't actually exist. Removed the field, its generation function, and now-unused imports.

## Test plan
- [x] `golangci-lint run` clean on touched packages
- [x] `go test ./internal/api/... ./internal/auth/... ./internal/webdav/...` — all pass (no existing webdav/auth test files)
- No behavior change intended beyond the CORS-credentials default; every other change is a drop-in comparison-function swap or dead-code removal.

### DIFF
--- a/internal/api/parsers.go
+++ b/internal/api/parsers.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"crypto/subtle"
 	"strconv"
 	"time"
 
@@ -95,7 +96,7 @@ func (s *Server) validateAPIKey(c *fiber.Ctx, apiKey string) bool {
 	if s.configManager != nil {
 		cfg := s.configManager.GetConfig()
 		if cfg.API.KeyOverride != "" && len(cfg.API.KeyOverride) == 32 {
-			if apiKey == cfg.API.KeyOverride {
+			if subtle.ConstantTimeCompare([]byte(apiKey), []byte(cfg.API.KeyOverride)) == 1 {
 				return true
 			}
 		}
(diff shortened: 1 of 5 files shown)


WhispersOfJ:

Issue #797, opened 2026-07-23
IsAdmin flag isn't enforced on any destructive/mutating route


WhispersOfJ:
Found during an independent source-level security audit. Filing as an issue rather than a PR since fixing this changes real authorization behavior and I'd rather confirm intended semantics first.

## What I found
`IsAdmin` is set on the first registered user (`internal/auth/service.go`'s `RegisterUser`) and returned in API responses, but grepping across `internal/api/*.go`, the only place it's actually consulted is `isAdminOrLoginDisabled` (`internal/api/auth_handlers.go`), which gates:
- whether `/auth/config` reveals registration-open state, and
- whether update/restart-triggering endpoints in `update_handlers.go` are reachable.

Every other mutating or destructive endpoint is reachable by **any authenticated user** regardless of `IsAdmin` — e.g. `/fuse/stop`, `/fuse/force-stop`, config `PUT`, bulk queue deletes, `/health/reset-all`, and a full system restart route.

## Why it (mostly) doesn't matter today, and why I'm still flagging it
`RegisterUser` hard-disables further registration after the first account, so in practice this is usually a single-user deployment and the gap has limited real-world blast radius. But the `IsAdmin` field, its DB column, and the frontend's apparent treatment of it as a real privilege boundary all imply a security control that isn't actually enforced anywhere that would matter (config changes, provider credential rotation, fuse mount teardown, full restart) — worth resolving one way or the other rather than leaving the model ambiguous, especially if multi-user support is ever revisited.

## Possible directions
1. Add a real `RequireAdmin` middleware and apply it to the config/system/fuse/queue-destructive route groups, so the existing `IsAdmin` field actually does something.
2. Or, if the intended model really is "every authenticated user is fully privileged, `IsAdmin` is informational only," document that explicitly (e.g. in the API docs / a comment on the field) so it doesn't read as an enforced boundary that isn't there.

Happy to send a PR for whichever direction matches the intended design.

javi11:
That's some sort of legacy stuff, there is no multiple users just 1 admin

WhispersOfJ:
Closing from our side - the stack this was filed against has moved off AltMount/BearMount entirely to a different Usenet client, so we're no longer tracking this here.

Issue #796, opened 2026-07-23
Unauthenticated SSRF via the SABnzbd ARR-credential auto-registration path


WhispersOfJ:
Found during an independent source-level security audit. Filing as an issue rather than a PR since a fix here changes real request-handling behavior and I'd rather get input on intended semantics first.

## Where
`internal/api/sabnzbd_handlers.go` — the `/sabnzbd` route accepts a second auth method beyond a real API key: `ma_username`/`ma_password` ("ARR credentials"), apparently meant to let Radarr/Sonarr self-register against AltMount. `ma_username` is treated as an arbitrary URL:

```go
arrURL, err := url.QueryUnescape(maUsername)
// ...
if instance := s.findARRInstanceByURL(arrURL); instance == nil {
    s.arrsService.RegisterInstance(c.Context(), arrURL, arrAPIKey)
}
```

`RegisterInstance` → `internal/arrs/instances/manager.go`'s `detectARRType` builds a `starr.Config{URL: arrURL, ...}` and issues real outbound HTTP requests trying the URL as Radarr, then Sonarr, then Lidarr, etc. — before any credential has actually been validated, and with no scheme/host allow-list (no rejection of loopback/link-local/RFC1918 targets).

`/sabnzbd` itself is also mounted on the root Fiber `app` ahead of the authenticated route group, so this whole path is reachable with zero authentication.

## Why it matters
An unauthenticated request with a crafted `ma_username` (e.g. a URL-encoded internal address, a metadata endpoint, an internal admin panel) makes AltMount issue outbound HTTP probes to an attacker-chosen destination as a side effect of "trying to authenticate" — classic SSRF. A successful-looking response from the probed host gets **persisted** as a new ARR instance with the attacker-supplied URL and API key, which is a real, non-blind side effect beyond just the request itself.

## Possible directions (would appreciate your take before I open a PR)
- Only allow ARR auto-registration via the already-authenticated API-key path, not the credential-guessing path.
- Or, if auto-registration must stay reachable this way, reject loopback/link-local/RFC1918 targets unless explicitly allow-listed (`net.ParseIP` + `IsLoopback`/`IsPrivate`/`IsLinkLocalUnicast`) before `detectARRType` ever dials out — though this would also block the very common case of Radarr/Sonarr actually running on an RFC1918 address, so it may need to be an explicit opt-in allow-list rather than a blanket block.

Happy to send a PR once there's agreement on which direction fits the intended design here.

PR #799, opened 2026-07-23
Security/leak fixes: consolidated (supersedes #792, #793, #794, #795)

WhispersOfJ:
## Summary

This consolidates four previously-separate PRs from an ongoing security/resource-leak audit of this repo into one branch, kept in sync going forward as the single source of truth on our fork's `main`:

- #792 — obscure the WebDAV password when creating the internal rclone mount config (already running in production as a local patched image)
- #793 — constant-time credential comparisons, CORS wildcard+credentials default, exact-match auth-skip, dead `DirectAuthSalt` removed
- #794 — archive path-traversal guard (RAR/7z internal entry names)
- #795 — 3 goroutine/fd-lifecycle leaks (health worker, queue manager, pool quota watcher)

Plus new work not yet in any prior PR, from two further audit passes:

- **Critical**: two independent path-traversal bugs in the *non-archive* NZB import path (NZB filename traversal in `parser.SanitizeNzbFilenames` and its sinks in `filesystem`/`singlefile`; a client-reachable `Category` field used raw as a directory-name component in four separate sinks, one of which — the symlink writer — is an arbitrary-file-delete-then-symlink primitive). Same bug class already fixed for archive-internal entries in #794, this closes the identical gap in the non-archive path.
- Prowlarr API-key exfiltration risk via indexer-supplied `DownloadURL` (no host-match check before attaching the key)
- 3 path-matching boundary bugs in `arrs/scanner/manager.go` (no separator boundary on prefix/contains checks, and a weak basename-only fallback that's the single source of truth for the Radarr delete-safety decision)
- A goroutine/fd leak on cancelling an in-progress NzbDav migration
- 2 slow unbounded-map leaks (a ghost-tracking map, a per-series episode-files cache)
- Real regression test coverage added for every fix above (none of the touched functions had traversal test cases before - only legitimate-input coverage existed)
- Two new CI workflows: a build/vet/lint/test gate (nothing previously ran `go vet` or golangci-lint), and a scheduled upstream-sync workflow so this fork's `main` keeps tracking `javi11/altmount:main` going forward, merging with `-X ours` so our fixes always win a conflicting hunk without blocking unrelated upstream changes

## Not included here (need your input on intended design, not just a code fix)

Two findings from the first audit pass were filed as issues instead of folded into a PR, since fixing them changes real request-handling/authorization behavior:
- #796 — unauthenticated SSRF via the SABnzbd ARR-credential auto-registration path
- #797 — `IsAdmin` isn't enforced on any destructive/mutating route

## Test plan
- [x] `go build ./...` / `go vet ./...` clean
- [x] `golangci-lint run` — 0 issues
- [x] `gopls check` — 0 diagnostics on every touched file
- [x] `go test ./...` — full suite passes, including new regression tests for every path-traversal fix
- [x] Full Docker image built and smoke-tested (container reports healthy, `/live` returns 200 with real JSON)


javi11:
Em dafuk, no way

PR #801, opened 2026-07-24
feat(nzbfilesystem): trigger immediate ARR repair search on playback holes

### LINKED ISSUE #798: [Feature request] Deal with missing articles during playback
Would love for Altmount to have the ability to quickly find a new version of a file that is being stream but is missing articles.

Say somebody is watching a video and during playback, as the read-ahead cache is being filled, missing articles are detected. Would be nice if Altmount could quickly search for (and replace) the current stream with a same quality version of the file without the current stream being impacted and the end user would be none the wiser.

WhispersOfJ:
## Summary

Closes #798.

A missing article confirmed during live playback (`holes.DecisionPad`) was previously only
recorded as `FILE_STATUS_DEGRADED`, with **no repair action at all** — the file's next
scheduled health check could be up to 90 days away for files older than 30 days
(`health.normalCheckInterval`), so the only real recovery path was a human noticing and
manually re-searching.

`padRecorder` now independently asks the owning ARR instance to blocklist+redownload the
title the moment a hole is confirmed, without touching the corrupted-file repair lifecycle:
no safety-folder move, no `FILE_STATUS_CORRUPTED`, no change to the existing DB status
semantics (stays `degraded`, still visible and streamable). The current stream is never
disturbed — ARR's own delete+search path only acts once it resolves a matching movie/episode
file via its existing path-matching guards (`radarrFileMatchesTarget` / its Sonarr
equivalent), which is independent of anything in AltMount's own mount.

Runs on its own goroutine off the recorder's single serialized worker loop, gated by the same
`"\x00degraded-pad"` debounce key already used for the DB status write, bounded to at most one
trigger per file per 30s debounce window regardless of how many holes accumulate in it.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./internal/nzbfilesystem/...` clean
- [x] `go test ./internal/nzbfilesystem/... -race` — full package suite passes, including 3
      new tests covering: library-path resolution, mount-path fallback resolution (file not
      yet relinked by an ARR), and nil-`arrs`/nil-`configGetter` no-op tolerance

### DIFF
--- a/internal/database/health_repository.go
+++ b/internal/database/health_repository.go
@@ -362,12 +362,104 @@ func (r *HealthRepository) GetFilesForRepairNotification(ctx context.Context, li
 		}
 		files = append(files, &health)
 	}
-
 	if err = rows.Err(); err != nil {
 		return nil, fmt.Errorf("failed to iterate files for repair notification: %w", err)
 	}
 
 	return files, nil
+}
+
+// GetFilesForImmediateRepair returns files with a pending immediate-repair
+// request (set by nzbfilesystem's padRecorder the moment a playback hole is
+// confirmed - see migration 035). The worker fires an ARR rescan for each
+// without a CheckFile pass or a status change; see
+// HealthWorker.triggerImmediateRepairOnly. Only file_path is needed here -
+// the worker re-fetches the full record per file (same pattern
+// GetFilesForRepairNotification's caller already uses) so a webhook-updated
+// library_path is never stale by the time the ARR trigger fires.
+func (r *HealthRepository) GetFilesForImmediateRepair(ctx context.Context, limit int) ([]*FileHealth, error) {
+	query := `
+		SELECT file_path
+		FROM file_health
+		WHERE immediate_repair_requested_at IS NOT NULL
+		ORDER BY immediate_repair_requested_at ASC
+		LIMIT ?
+	`
+
+	rows, err := r.db.QueryContext(ctx, query, limit)
+	if err != nil {
+		return nil, fmt.Errorf("failed to query files for immediate repair: %w", err)
+	}
+	defer rows.Close()
+
+	var immediateFiles []*FileHealth
+	for rows.Next() {
+		var filePath string
+		if err := rows.Scan(&filePath); err != nil {
+			return nil, fmt.Errorf("failed to scan file path for immediate repair: %w", err)
+		}
+		immediateFiles = append(immediateFiles, &FileHealth{FilePath: filePath})
+	}
+	if err = rows.Err(); err != nil {
+		return nil, fmt.Errorf("failed to iterate files for immediate repair: %w", err)
+	}
+
+	return immediateFiles, nil
+}
+
+// SetImmediateRepairRequested marks filePath as needing an immediate ARR
+// repair trigger on the health worker's next cycle. Idempotent: repeated
+// calls before the worker processes it just refresh the timestamp, which is
+// harmless since GetFilesForImmediateRepair only cares that it's non-NULL.
+// A no-op (not an error) if the health record doesn't exist yet.
+func (r *HealthRepository) SetImmediateRepairRequested(ctx context.Context, filePath string) error {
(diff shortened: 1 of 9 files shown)


javi11:
### Alternative: route the rescan through the health queue instead of triggering it directly

Right now `padRecorder.triggerImmediateRepairSearch` calls `arrs.TriggerFileRescan(...)` directly on its own goroutine. It works, but it bypasses every safeguard the health worker already enforces around ARR triggers:

- **No `GetRepairEnabled()` gate** — it fires an ARR search even when repair is disabled.
- **No repair-budget accounting** (`RepairRetryCount` vs `MaxRepairRetries`) — the same permanently-broken release can be re-downloaded on every 30s debounce window, forever. The worker caps this (`worker.go:520`).
- **No back-off / indexer failure logging** — the worker's repair path has exponential cooldown and `LogIndexerImport`.

**Proposal (Option B):** instead of triggering directly, have the pad recorder enqueue a lightweight "repair-requested" state into the health queue, and let the health worker own the ARR trigger — the way it already owns every other ARR rescan (single serialized worker).

Crucially this must preserve the PR's core invariant (keep the current stream alive): the worker branch for this state must **trigger the ARR rescan only** — skip `CheckFile` (playback already confirmed the hole; re-verifying is redundant), skip `moveMetadataToSafetyFolder`, and leave the file `degraded`/`FILE_STATUS_DEGRADED` and streamable. So it can't reuse `triggerFileRepair` as-is (that always moves metadata + condemns the file); it needs a dedicated non-destructive branch.

**Why this is better than the direct call:** the ARR trigger lives in one place, and it automatically inherits the `repairEnabled` gate, the repair-retry budget (so we don't re-download the same broken release indefinitely), back-off, and indexer logging.

**What to avoid:** simply scheduling an immediate *normal* re-check (`scheduled_check_at = now`, status `pending`) is not enough — the checker would either re-classify as `Degraded` (no repair → original bug returns) or as `Corrupted` (safety-folder move + `FILE_STATUS_CORRUPTED` → breaks the live stream). Neither matches the intent, so a dedicated repair-requested state is required.

**Cost:** a new health status (or a `repair_requested` flag) + a dedicated non-destructive branch in the worker's dispatch. Moderate change, but it centralizes ARR triggering and closes the safeguard gaps above.

WhispersOfJ:
Pushed a change implementing Option B from the review above.

`padRecorder` no longer calls `arrs.TriggerFileRescan` directly. It now just sets a lightweight
`immediate_repair_requested_at` flag on the file's health record (migration 035 — a nullable
timestamp column rather than a new `HealthStatus` value, since `status` is CHECK-constrained and
a new enum value would need a full table rebuild; a plain `ADD COLUMN` needs none).

The health worker's dispatch cycle (`runHealthCheckCycle`) now has a third lane alongside the
existing scheduled-check and repair-notification lanes: it picks up files with a pending
`immediate_repair_requested_at`, re-fetches the current record (so a concurrent webhook-updated
`library_path` is never stale), and fires the ARR rescan via a new `triggerImmediateRepairOnly`.
That function owns everything the direct call skipped:

- `GetRepairEnabled()` gate — skipped and cleared (not left stuck) when repair is disabled
- `RepairRetryCount` / `MaxRepairRetries` budget, same as every other repair trigger
- indexer failure logging on a failed `TriggerFileRescan` call
- clears the flag and increments the retry counter after firing, regardless of outcome

Deliberately still does **not** call `CheckFile` and does **not** move metadata to the
corrupted-safety folder or change `FILE_STATUS` — the file stays `degraded`/streamable the whole
time, since a live stream may still be reading it when the pad fires.

Verified: `go build ./...`, `go vet ./...`, and `go test -race ./...` all clean across the whole
repo (not just the touched packages).

WhispersOfJ:
Closing - the stack this fork supported has moved off AltMount/BearMount entirely to a different Usenet client (nzbdav/nzbdav), so this fork is being retired and this feature won't be carried forward from our side.

PR #1, opened 2025-09-04
ci: publish multi‑arch manifests with buildx imagetools and lowercase image name

JeffreyBytes:
## Summary

- Replace docker manifest create/push with docker buildx imagetools create to assemble multi‑arch manifests from per‑arch tags.
- Normalize image name to lowercase for GHCR compliance (fork owners often have mixed‑case usernames).
- Keep existing tag scheme: vX.Y.Z, X.Y.Z, vX, X, vX.Y, X.Y, latest.

### Why

- Previous workflow failed at manifest creation because buildx pushes per‑arch tags (e.g., :…‑amd64) as OCI manifest lists; docker manifest create cannot nest lists (“… is a manifest list”).
- GHCR requires lowercase image references; mixed‑case org/repo names cause “repository name must be lowercase”.

## What changed

- Create manifests with docker buildx imagetools create:
  - Sources: ghcr.io/:-amd64 and -arm64.
  - Targets: vX.Y.Z and X.Y.Z; vX and X; vX.Y and X.Y; latest.
- Add Buildx setup to the create‑manifest job.
- Normalize `IMAGE_NAME` to lowercase at runtime and use it for:
    - docker/metadata‑action images
    - imagetools target/source references

## Notes

- No change to image content; only how multi‑arch manifests are created and tagged.
- Still supports pushing single‑arch tags (…‑amd64, …‑arm64).
- Uses default `GITHUB_TOKEN` to authenticate to GHCR; requires “Workflow permissions: Read and write”.

## Validation

- Tested in fork with tags `v0.0.1‑test1` and `v0.0.1‑test2`:
    - amd64 and arm64 images built and pushed
    - imagetools successfully created multi‑arch manifests for all tag variants
    - Lowercasing fix resolved GHCR “repository name must be lowercase” error

## Checklist

- [x] Builds: pass in CI on PR (unit tests + lint).
- [x] Release workflow: verified in fork on tag push.
- [x] Backward compatible: yes (tag scheme preserved; no runtime behavior change).

PR #806, opened 2026-07-26
fix: close sync.Cond lost-wakeup race and bound import budget wait

### LINKED ISSUE #805: Recurring FUSE read hang (D-state, zero connections) during sustained concurrent import load
## Summary

Recurring, reproducible FUSE read hangs on the import staging path during Radarr/Sonarr's post-download `ffprobe` validation pass. A read gets permanently wedged in D-state (uninterruptible sleep) with **zero** NNTP connections open and **zero** log output for the affected file — it never errors, never times out, and blocks the reading process (and sometimes the whole container's shutdown) indefinitely. Recovery required restarting the consumer; several times the consumer's own PID1 got wedged waiting on the D-state child, requiring the mount owner itself to be recreated to release it.

Hit this 8+ times in a single ~3 hour session, always a different file, no pattern by release group/indexer/quality tier (REMUX, WEB, and raw multi-file BDMV segments all affected).

## Root cause(s) found — two independent bugs

Filed together with a PR fixing both. See the PR for the actual diffs; this issue documents the investigation and evidence.

### 1. Lost `sync.Cond` wakeup in `UsenetReader`

`internal/usenet/usenet_reader.go`'s download-scheduling loop (`downloadManager`) throttles prefetch using a `sync.Cond` (`b.cond`, tied to `b.mu`): it holds `b.mu`, checks `ahead := b.nextToDownload - b.rg.GetCurrentIndex()`, and calls `b.cond.Wait()` if over the prefetch limit. The wakeup call sites — `Read()`'s post-`rg.Next()` signal, and each download goroutine's completion signal — both called `b.cond.Signal()` **without holding `b.mu`**. Critically, `segmentRange` (`b.rg`) has its **own separate `sync.RWMutex`**, so the state `downloadManager` checks isn't actually guarded by the same lock the condition variable is tied to.

Race: `downloadManager` reads a stale `currentRead`, decides to wait, and is about to call `Wait()` — but `Read()` concurrently advances the index and signals in that exact gap. `sync.Cond` does not queue signals; with no goroutine yet parked in `Wait()`, the signal is silently lost. `downloadManager` then calls `Wait()` and blocks forever — no more segments get scheduled, `Read()`'s next call blocks on data that will never arrive.

Deployed alone first to confirm/isolate — **did not stop the hang recurring** (observed directly: same D-state signature reappeared on a container confirmed running this fix, ~6 minutes after deploy). Real bug, worth fixing, but not the whole story.

### 2. Unbounded wait on the import connection budget

`downloadSegmentWithRetry`'s `b.budget.AcquireImportConnection(ctx)` call — the global import-connection semaphore — runs **before any per-attempt timeout exists**, using the reader's own long-lived context. Under heavy concurrent import load (multiple readers doing large `ffprobe`-style probes simultaneously — our exact reproduction conditions), if the queue backs up for any reason, a waiter near the back can legitimately wait an unbounded amount of time — indistinguishable from a permanent hang from the outside, and explains why only a restart (which discards the whole queue) ever "fixed" it rather than it resolving on its own.

Deployed on top of fix #1 (same build) — held clean for 12+ consecutive checks under continued real concurrent import load, where fix #1 alone had already failed to hold for even one check.

### Related, not fixed here

While tracing #2, found a related gap in the vendored `nntppool` client: `tryGroup`'s per-attempt timeout is a one-shot `time.Timer` — once it fires it cannot fire again (the code's own comment acknowledges this), but the "reader already committed" branch falls through and keeps looping past that point with no working stall-detection for a stream that stalls *after* it starts. Traced whether the caller's own context deadline still rescues this via propagation (`reqCtx.Done()` is one of the select's cases) — it does, so this alone doesn't explain multi-*minute* hangs, but it's a real gap worth closing independently. A genuine idle-stall watchdog would need progress tracking (last-byte-received timestamp) that doesn't currently exist on the `Request`/`attemptState` model — would require instrumenting the actual socket-read loop, out of scope for the linked PR.

## Evidence

- BearMount's own dashboard showed affected downloads as `STALLED`, `0 B/s`, stuck at a fixed ~960KB offset regardless of file size (observed identically on a 2.89GB file and a 62.73GB file simultaneously).
- `ss -tn | grep :563` (provider port) → 0 established connections during a hang.
- `docker exec <consumer> cat /proc/<pid>/stat` → process state `D`, persisting across repeated checks (same PID) for 5+ minutes until the consumer was restarted.
- No log line referencing the affected file during the hang window on the pre-fix build — silent, not failing (part of why this took hours to trace; fix #2 adds a log line on the timeout path specifically to close this gap for next time).
- The affected file's own health record showed `status: healthy` from its last small/sampled check — the sampled path succeeds while a full/large read on the same file hangs.
- Server-side context at the time: a separate, unrelated finding worth being aware of when reproducing — the configured NNTP provider was independently running a high missing-article rate (own log warning: "NNTP provider has high missing article rate — consider using a backup provider", ~700/min at the time), which explains a portion of *failed* imports in the same session but is orthogonal to the *hung* (never-erroring) reads this issue is about.

## Verification

- Both fixes: clean `go build`, full `internal/usenet` package test suite passes with `-race` (including pre-existing sequential/prefetch/storm tests written for this exact area).
- Deployed live against a real, heavily-loaded stack (concurrent import batches from two consumers, hundreds of queued items) for validation, not just unit tests.

## Reproduce

1. Run sustained, concurrent import activity from multiple consumers against an instance with default settings (`mount_type: fuse`, default import connection budget).
2. Given enough large-file sequential reads over ~10-30 minutes under real concurrent load, one will eventually hang in D-state instead of completing normally.
3. Check `ps aux | grep <reader process>` + `/proc/<pid>/stat` on the consumer to confirm state `D`; check `ss -tn` on the server process to confirm zero active provider connections during the hang.

## Environment

- `mount_type: fuse`, hanwen FUSE backend
- Single NNTP provider, `max_connections: 25`
- `vfs_cache_mode: full`
- Docker, `cap_add: SYS_ADMIN`, `/dev/fuse`

WhispersOfJ:
## Summary

Fixes two independent bugs behind a recurring FUSE read hang (see linked issue for the full investigation and evidence): a lost `sync.Cond` wakeup in `UsenetReader`'s download-scheduling loop, and an unbounded wait on the import connection budget semaphore. Both were required — the first alone did not stop the hang recurring in live testing; the second, deployed on top of the first, held clean under sustained concurrent import load where the first alone had already failed.

## Changes

**`internal/usenet/usenet_reader.go`**

1. `Read()`'s post-`rg.Next()` wakeup and the download goroutine's completion signal now hold `b.mu` around `b.cond.Signal()`. Without this, `downloadManager`'s own check-then-`Wait()` on the prefetch-ahead condition (also guarded by `b.mu`) could race against a signal fired while the state it depends on (`segmentRange.GetCurrentIndex()`, guarded by a *different*, separate mutex) had just changed — `sync.Cond` doesn't queue signals, so a signal landing in that gap is silently lost and the scheduler blocks forever.

2. `downloadSegmentWithRetry`'s `b.budget.AcquireImportConnection(ctx)` call is now bounded by a 5-minute `context.WithTimeout`, with a warning log on timeout. Previously this ran on the reader's own long-lived context with no timeout at all, before any per-attempt deadline existed — under heavy concurrent import load a queued waiter could wait indefinitely with zero log output, indistinguishable from a permanent hang. 5 minutes is deliberately generous: the original unbounded design was intentional (comment: "queue wait never burns the fetch deadline"), so this preserves normal heavy-contention behavior and only fires for genuinely pathological queueing.

## Testing

- `go build ./...` clean.
- `go test -race ./internal/usenet/...` — full package passes, including the pre-existing sequential/prefetch/storm tests already covering this area.
- Deployed live against a real, heavily-loaded stack (concurrent import batches from two consumers, hundreds of queued items over multiple hours) rather than relying on unit tests alone, since this is a timing-dependent race that unit tests may not reliably reproduce. Fix #1 alone was deployed and observed to still hang (~6 min after deploy, confirmed via `docker inspect` that the hung container was genuinely running the fix-#1-only build). Fix #2 on top of #1 held clean for 12+ consecutive checks under continued real load.

## Not included

Traced a related gap in `tryGroup`'s per-attempt timer (one-shot `time.Timer`, can't re-fire after a stream stalls post-commit) while investigating — the caller's own context deadline still rescues that case via propagation, so it wasn't the cause of *this* symptom, but it's a real gap worth closing separately. Left out of this PR since a correct fix needs progress tracking that doesn't exist on the current `Request`/`attemptState` model (would mean instrumenting the actual socket-read loop) — happy to open a follow-up issue/PR for it if useful.

Fixes #805


WhispersOfJ:
Closing - the stack this fork supported has moved off AltMount/BearMount entirely to a different Usenet client (nzbdav/nzbdav), so this fork is being retired and this fix won't be carried forward from our side. Thanks for the review time either way.

PR #815, opened 2026-08-07
fix(metadata): preserve ModifiedAt across updates and stabilise file & directory mtime

DarkseidAM:
## Problem

Every time Altmount restarts, its health worker fires `EventTypeFileHealthy` for all known-healthy files immediately. Each event calls `UpdateFileStatus` → `UpdateFileMetadata` → `WriteFileMetadata`, which does an atomic `os.Rename`. That rename always sets the `.meta` disk mtime to `now()`, even when the file's logical `ModifiedAt` hasn't changed at all.

For setups where Altmount's WebDAV is mounted via rclone (or any client that reads `Last-Modified`), this causes media servers (Jellyfin, Plex, Emby, etc.) to detect a timestamp change on every container restart and re-probe every affected file over the mount — wasting bandwidth, CPU, and blocking concurrent imports.

## Root Cause

1. **`UpdateFileMetadata` — no enforcement of ModifiedAt immutability**: Any callback passed to `UpdateFileMetadata` could accidentally alter `ModifiedAt`.
2. **`WriteFileMetadata` — `os.Rename` silently updates disk mtime**: After `os.Rename(tmpPath, metadataPath)`, the destination file gets `mtime = now()`.
3. **Virtual Directory `modTime` churn**: Virtual directories previously returned `time.Now()` on every stat, causing folder-level timestamp churn on media scanners.

### 1. `UpdateFileMetadata` — unconditional snapshot + restore & clear contract
- Snapshot `ModifiedAt` before invoking the callback and restore it unconditionally afterwards.
- Documented that `UpdateFileMetadata` is strictly reserved for non-content RMW operations (health status, known-holes). Content-modifying paths must use `WriteFileMetadata` / `CreateFileMetadata` directly.

### 2. `WriteFileMetadata` — `os.Chtimes` after rename
- Pin the `.meta` disk mtime back to `ModifiedAt` immediately after the atomic rename:
  ```go
  if metadata.ModifiedAt != 0 {
      t := time.Unix(metadata.ModifiedAt, 0)
      _ = os.Chtimes(metadataPath, t, t)
  }
  ```

### 3. Virtual Directory `modTime` — `max(children's ModifiedAt)`
- Virtual directory `Stat` and `Readdir` paths now derive `modTime` dynamically via `max(children's ModifiedAt)` (using `GetDirectoryModTime()`).
- Stays frozen across restarts (because health sweeps never touch any child's `ModifiedAt`), while advancing cleanly when new files are actually imported. Static epoch (`2000-01-01`) is used purely as a fallback for empty directories.

## Testing

- Added `TestGetDirectoryModTime` verifying empty directories, multi-file max aggregation, stability across status updates, and advancement on new file creation.
- Verified on production (rclone WebDAV sidecar + Jellyfin 10.11): `.meta` disk mtimes and directory timestamps remain stable across container restarts and continuous health sweeps.

### DIFF
--- a/internal/nzbfilesystem/metadata_remote_file.go
+++ b/internal/nzbfilesystem/metadata_remote_file.go
@@ -508,11 +508,15 @@ func (mrf *MetadataRemoteFile) Stat(ctx context.Context, name string) (bool, fs.
 
 	// Check if this is a directory first
 	if mrf.metadataService.DirectoryExists(normalizedName) {
+		modT := mrf.metadataService.GetDirectoryModTime(normalizedName)
+		if modT.IsZero() {
+			modT = staticVirtualDirModTime
+		}
 		info := &MetadataFileInfo{
 			name:    filepath.Base(normalizedName),
 			size:    0,
 			mode:    os.ModeDir | 0755,
-			modTime: staticVirtualDirModTime, // Fixed static time for virtual directories to prevent media server re-scanning
+			modTime: modT,
 			isDir:   true,
 		}
 		return true, info, nil
@@ -669,7 +673,18 @@ func (mvd *MetadataVirtualDirectory) Readdir(count int) ([]fs.FileInfo, error) {
 
 	// Add directories first
 	for _, dirInfo := range dirInfos {
-		infos = append(infos, dirInfo)
+		subPath := filepath.Join(mvd.normalizedPath, dirInfo.Name())
+		subModT := mvd.metadataService.GetDirectoryModTime(subPath)
+		if subModT.IsZero() {
+			subModT = staticVirtualDirModTime
+		}
+		infos = append(infos, &MetadataFileInfo{
+			name:    dirInfo.Name(),
+			size:    0,
+			mode:    os.ModeDir | 0755,
+			modTime: subModT,
+			isDir:   true,
+		})
 		if count > 0 && len(infos) >= count {
 			return infos, nil
 		}
@@ -734,11 +749,15 @@ func (mvd *MetadataVirtualDirectory) Readdirnames(n int) ([]string, error) {
 
 // Stat implements afero.File.Stat
 func (mvd *MetadataVirtualDirectory) Stat() (fs.FileInfo, error) {
+	modT := mvd.metadataService.GetDirectoryModTime(mvd.normalizedPath)
+	if modT.IsZero() {
+		modT = staticVirtualDirModTime
+	}
 	info := &MetadataFileInfo{
 		name:    filepath.Base(mvd.normalizedPath),
 		size:    0,
 		mode:    os.ModeDir | 0755,
-		modTime: staticVirtualDirModTime,
+		modTime: modT,
 		isDir:   true,
 	}
 	return info, nil
--- a/internal/nzbfilesystem/metadata_remote_file.go
+++ b/internal/nzbfilesystem/metadata_remote_file.go
@@ -33,13 +33,22 @@ import (
 	"github.com/spf13/afero"
 )
 
-var staticVirtualDirModTime = time.Date(2000, 1, 1, 0, 0, 0, 0, time.UTC)
+// epochFallbackModTime is reported when no real timestamp is available (a file
+// whose ModifiedAt was never set, or a directory that is not backed by an
+// on-disk metadata dir). A fixed date is deliberate: returning time.Now() would
+// make every stat look like a fresh modification and force media scanners to
+// re-probe on every call.
+var epochFallbackModTime = time.Date(2000, 1, 1, 0, 0, 0, 0, time.UTC)
 
+// getFileModTime maps a proto ModifiedAt to the mtime reported over WebDAV and
+// FUSE. Note that ReadFileMetadataLite treats modified_at as best-effort — it
+// stays zero when the field sits past liteScanBytes — so listings can legitimately
+// fall back to the epoch here.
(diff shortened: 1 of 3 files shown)


internal/nzbfilesystem/metadata_remote_file.go
```diff
@@ -987,6 +1001,51 @@ func (idx *segmentOffsetIndex) getOffsetForSegment(segmentIndex int) int64 {
 	return idx.offsets[segmentIndex]
 }
 
+func (mvf *MetadataVirtualFile) startIdleWatchdog(idleTimeout time.Duration) {
```
javi11: What is this for?

DarkseidAM: Removed in force-push. This was mistakenly included from PR #816; PR #815 now strictly contains the ModifiedAt immutability and directory modTime fix.

internal/nzbfilesystem/metadata_remote_file.go
```diff
@@ -503,7 +512,7 @@ func (mrf *MetadataRemoteFile) Stat(ctx context.Context, name string) (bool, fs.
 			name:    filepath.Base(normalizedName),
 			size:    0,
 			mode:    os.ModeDir | 0755,
-			modTime: time.Now(), // Use current time for directories
+			modTime: staticVirtualDirModTime, // Fixed static time for virtual directories to prevent media server re-scanning
```
javi11: If you do this, rclone will never refresh directory even if the content inside changed, you should derive this from the real meta directory

DarkseidAM: Altmount uses active VFS notifications via `rcloneClient.RefreshDir` (`vfs/refresh` RC API) whenever files are imported or modified (see `vfs_notifier.go`). Because directory cache updates are explicitly pushed to rclone on changes, rclone does not rely on passive directory `modTime` polling. Freezing virtual directory `modTime` prevents media scanners (Plex/Jellyfin) from re-probing virtual directories on container restarts.

javi11:
Two design concerns on this fix that I think are worth addressing before merge:

**1. `UpdateFileMetadata`'s unconditional `ModifiedAt` restore is a footgun for future callers.**
The snapshot/restore in `UpdateFileMetadata` (service.go) now silently discards any `ModifiedAt` change made inside the callback, with no error or log. Today's two callers (`UpdateFileStatus`, `AddKnownHoles`) are correctly non-content updates, but the function still reads as a general-purpose read-modify-write helper — a future contributor doing a legitimate "content actually changed, bump `ModifiedAt`" update would naturally reach for this same pattern and have it quietly reverted. Suggest either:
- Documenting/enforcing that content-modifying paths must go through `WriteFileMetadata`/`CreateFileMetadata` directly (which don't touch `ModifiedAt` — caller sets it explicitly), and treating `UpdateFileMetadata` as reserved for non-content RMW only, or
- Adding an explicit opt-in (e.g. a bool param or a separate `UpdateFileMetadataAllowModifiedAt`) for the rare legitimate case, so it fails loud/obvious rather than silently.

**2. Freezing directory mtime to a static constant trades one bug for another.**
`dir_cache_time` (rclone VFS) and FUSE `attr_timeout`/`entry_timeout` are both plain TTL-based invalidation, not mtime-diff-based, so this doesn't break AltMount's own mount/FUSE freshness. But the reason this PR exists is that *external* tools (Jellyfin) read WebDAV `Last-Modified` on directories as a change signal — many incremental scanners use "did this folder's mtime change?" to skip walking unchanged folders. If every directory now always reports the same static `2000-01-01`, those tools lose any signal that a folder just got a new file added, and could skip re-scanning it indefinitely — a silent missing-content bug, arguably worse than the noisy-but-harmless restart storm this PR fixes.

Suggest computing directory mtime as `max(children's ModifiedAt)` instead of a hardcoded constant. That still solves the original problem — a health-check restart doesn't touch any child's `ModifiedAt`, so the directory mtime stays stable across restarts — while correctly advancing when a file is actually added/modified/removed, at the cost of a cheap aggregation over `Readdir`'s already-iterated results.

internal/nzbfilesystem/metadata_remote_file.go
javi11: Yes it does not, but that refresh fails sometimes

DarkseidAM: Updated in `6e949d4`:

1. Added `GetDirectoryModTime(virtualPath)` to compute `max(children's ModifiedAt)` for virtual directory `Stat` and `Readdir` paths (with epoch fallback only for empty/unresolvable dirs).
2. Clarified `UpdateFileMetadata` docstring to make its non-content RMW contract explicit and point content-modifying paths directly to `WriteFileMetadata` / `CreateFileMetadata`.

DarkseidAM:
Addressed both points in commit `6e949d4`:

1. **`UpdateFileMetadata` Contract**: Updated the docstring to clearly document that `UpdateFileMetadata` is strictly reserved for non-content RMW operations (where `ModifiedAt` is preserved), and that any callers intending to legitimately modify content timestamps must use `WriteFileMetadata` / `CreateFileMetadata` directly.
2. **Directory `modTime` Calculation**: Added `GetDirectoryModTime(virtualPath)` to `MetadataService` to dynamically aggregate `max(children's ModifiedAt)` across directory `Stat` and `Readdir` paths, using the static epoch timestamp purely as a fallback for empty directories.

javi11 (approved):

PR #816, opened 2026-08-09
fix(nzbfs): bind file context to HTTP request context and cancel on close

DarkseidAM:
### Description

This PR ensures that virtual file handles opened via WebDAV/HTTP bind their execution context to the caller's request context and cancel in-flight reader downloads when closed or disconnected.

### Changes

- **Request-bound File Context**: Wrapped `OpenFile` context in a child context derived from the HTTP request context (`fileCtx, cancel := context.WithCancel(ctx)`), and explicitly invoke `mvf.cancelCtx()` in `MetadataVirtualFile.Close()`. This ensures client disconnects or handle closures immediately cancel in-flight Usenet reader downloads.

### DIFF
--- a/cmd/altmount/cmd/setup.go
+++ b/cmd/altmount/cmd/setup.go
@@ -521,7 +521,7 @@ func createHTTPServer(apiServer *api.Server, app *fiber.App, webdavHandler *webd
 		Addr:         fmt.Sprintf(":%d", port),
 		Handler:      mainHandler,
 		IdleTimeout:  time.Minute * 5,
-		WriteTimeout: time.Minute * 3,
+		WriteTimeout: time.Minute * 30,
 		ReadTimeout:  time.Minute * 5,
 	}
 }
--- a/cmd/altmount/cmd/setup.go
+++ b/cmd/altmount/cmd/setup.go
@@ -521,7 +521,7 @@ func createHTTPServer(apiServer *api.Server, app *fiber.App, webdavHandler *webd
 		Addr:         fmt.Sprintf(":%d", port),
 		Handler:      mainHandler,
 		IdleTimeout:  time.Minute * 5,
-		WriteTimeout: time.Minute * 30,
+		WriteTimeout: time.Minute * 3,
 		ReadTimeout:  time.Minute * 5,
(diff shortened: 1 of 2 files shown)


javi11:
I think this can be achieve by seting up a context with cancel timeout on open the file, you don't need any watchdog

DarkseidAM:
Refactored as suggested in the force-push: removed the per-file `startIdleWatchdog` goroutine entirely.

The PR now relies on:
1. `context.WithCancel(r.Context())` bound to each `OpenFile` handle so HTTP client disconnects immediately propagate cancellation to in-flight NNTP segment downloads.
2. Reduced HTTP server `WriteTimeout` (3 minutes) in `setup.go` to ensure idle/stalled HTTP response streams auto-close and cancel their request context.

javi11:
Now you add this cancelCtx but you don't use it anywhere

DarkseidAM:
Apologies, forgot to stage the commit
Updated `MetadataVirtualFile.Close()` to invoke `mvf.cancelCtx()` when the file handle is closed.

javi11:
## Review

Change #1 (context binding in `metadata_remote_file.go`) makes sense — verified the full call chain, it correctly ties the file's lifetime to the HTTP request and closes a real race window on FUSE where a reader could otherwise inherit an uncancelable context.

Change #2 (`WriteTimeout` 30m → 3m in `setup.go`) — **drop this part.** `WriteTimeout` isn't an idle/stall timeout; it's a hard cap on the total request-response duration that never resets on activity. It would forcibly cut off large un-ranged WebDAV downloads and kill the app's own long-lived SSE endpoints (`/api/queue/stream`, `/api/health/stream`) every 3 minutes regardless of their keep-alive traffic. Change #1 already delivers the actual disconnect/stall detection this PR wants, so the `WriteTimeout` change isn't needed and introduces regressions.

DarkseidAM:
Dropped the `WriteTimeout` change in commit `883f2ae`. The PR now strictly contains the request context binding and cancellation on `MetadataVirtualFile.Close()`.

javi11 (approved):

PR #836, opened 2026-08-23
fix(database): normalize virtual health paths

dclipca:
## Summary

- canonicalize health and import-history virtual paths to forward-slash, root-relative form at every write/read boundary
- add SQLite and PostgreSQL migration 035 to normalize legacy rows and conservatively merge file_health collisions without discarding repair evidence
- keep dialect-specific SQL at the existing database boundary and add migration/repository regression coverage

## Why

Legacy leading separators and Windows separators can make logically identical paths miss health/history joins or collide under the file_health unique key. Normalizing only new writes leaves existing rows inconsistent, so the migration and repository behavior land together.

This is separate from #667: that PR fixed failed-NZB filesystem namespacing; this change repairs persisted virtual-path identity.

## Validation

- go test ./internal/database
- go vet ./internal/database/...
- SQLite migration/repository regression tests
- PostgreSQL migration smoke test
- git diff --check

The migration is intentionally non-reversible because collision merging cannot be losslessly undone.


internal/database/migrations/sqlite/035_normalize_virtual_paths.sql
```diff
@@ -0,0 +1,122 @@
...
+--
+-- Keep the two virtual-path columns in one representation: forward slashes and
+-- no leading/trailing separators. import_history has no path uniqueness constraint, so it
+-- can be normalized in place. file_health is UNIQUE(file_path), so it is rebuilt
+-- through a collision-free aggregate before the old rows are replaced.
+UPDATE import_history
+SET virtual_path = trim(replace(virtual_path, char(92), '/'), '/')
+WHERE virtual_path <> trim(replace(virtual_path, char(92), '/'), '/');
+
+DROP TABLE IF EXISTS file_health_path_normalized;
```
javi11: This would drop a real table of that name if one ever existed. Doesn't buy anything either — temp tables die with the session.

internal/database/migrations/sqlite/035_normalize_virtual_paths.sql
```diff
@@ -0,0 +1,122 @@
...
+    SELECT h.*,
+        ROW_NUMBER() OVER (
+            PARTITION BY canonical_path
+            ORDER BY CASE status
+                WHEN 'corrupted' THEN 6
+                WHEN 'repair_triggered' THEN 5
+                WHEN 'degraded' THEN 4
+                WHEN 'checking' THEN 3
+                WHEN 'pending' THEN 2
+                WHEN 'partial' THEN 2
```
javi11: Dead code — `partial` was dropped from the CHECK constraint back in 004, so no row can hold it. Same on line 76. Reads like live logic.

internal/database/migrations/sqlite/035_normalize_virtual_paths.sql
```diff
@@ -0,0 +1,122 @@
...
+            PARTITION BY canonical_path
+            ORDER BY CASE WHEN NULLIF(download_id, '') IS NOT NULL THEN 0 ELSE 1 END,
+                updated_at DESC, id DESC) AS download_id_rank
+    FROM file_health_path_normalized h
+)
+SELECT
+    COALESCE(MIN(CASE WHEN file_path = canonical_path THEN id END), MIN(id)) AS id,
+    canonical_path AS file_path,
+    MAX(CASE WHEN library_rank = 1 AND NULLIF(library_path, '') IS NOT NULL THEN library_path END) AS library_path,
+    MAX(CASE WHEN status_rank = 1 THEN CASE status WHEN 'partial' THEN 'corrupted' ELSE status END END) AS status,
```
javi11: The merge is column-wise across 9 independent rank windows, so the survivor is a composite that existed in no prior state — status from one row, counters from another, timestamps from a third. `MAX(retry_count)` in particular trips the repair gate at `worker.go:523` a cycle early.

Not strictly wrong, "worst-of wins" is a fair reading of not discarding evidence, and it only hits real collisions. But picking one winner row and backfilling only its NULL/empty columns would be easier to reason about. At minimum worth a comment saying this is deliberate.

javi11 (commented):
Nice win on the join change — dropping `TRIM()` from the join predicates makes them sargable so `idx_file_health_path` actually gets used. That part I like.

Migration 035 is the blocker though. It does `DELETE FROM file_health` + full reinsert, so it rewrites the whole table and all 8 indexes regardless of how many rows actually need fixing. I benchmarked it on SQLite/WAL, NVMe:

| rows | dirty | 035 |
|---|---|---|
| 50k | 0 | 1.6s |
| 200k | 0 | 9.1s |
| 200k | 1/1000 | 9.6s |

Cost scales with library size, not with rows needing repair — a clean 200k table pays the same 9.1s. That's a fast dev machine; most people run AltMount on a NAS or Pi, so 500k–1M rows is minutes of silent hang at startup with no logging. Users will assume it's stuck and kill it mid-migration.

Doing only the dirty rows is 75ms at 200k (~120x), and the merge then runs over 200 rows instead of 200k:

```sql
UPDATE file_health SET file_path = trim(replace(file_path, char(92), '/'), '/')
WHERE file_path <> trim(replace(file_path, char(92), '/'), '/')
  AND NOT EXISTS (SELECT 1 FROM file_health c
                  WHERE c.file_path = trim(replace(file_health.file_path, char(92), '/'), '/'));
```

Whatever is still non-canonical after that is exactly the collision set — run the existing merge over just those.

Two more things on the migration: it materializes two full copies of the table in temp storage, and in Docker SQLite's temp store often lands on a small `/tmp` or a RAM-backed tmpfs, so a big library can hit `database or disk is full` or OOM the container. And on Postgres the delete+reinsert leaves 100% dead tuples — table and indexes stay at double size until autovacuum catches up, and you can't `VACUUM` inside the goose transaction.

---

Two things outside the diff that belong in this PR:

**`internal/database/health_repository.go:1996` (`GetFilesByPaths`)** — this one got missed, still `strings.TrimPrefix(path, "/")`, the same idiom you replaced in `DeleteHealthRecordsBulk` just above. No backslash conversion, no trailing trim, no double-slash. Matters more now that the joins lost their `TRIM()` — that was what was covering for it. Caller is `POST /health/regenerate-symlinks`, which takes `file_paths` straight from the request, so `\movies\x.mkv` silently matches nothing and it reports 0 files.

**`health_repository.go:1731`** — while you're on that line, the prefix is `+ "%"` but every other prefix query in the file uses `+ "/%"` (569, 605, 635). So `tv/Show` also matches `tv/ShowOther/...`, and this is a DELETE. Pre-existing, but `TrimPrefix` → `Trim` now also strips a trailing separator a caller might have passed as the bound.

PR #878, opened 2026-08-31
fix(metadata): normalize backslashes in virtual paths to prevent FUSE open() deadlock

### LINKED ISSUE #660: FUSE open() on filename containing backslash deadlocks indefinitely in invalidate_inode_pages2

firestaerter3:
Fixes #660. Supersedes #685, which could not be reopened (GitHub returns "the pull
request cannot be reopened"), so this is the same branch with the conflict resolved.

### The bug

A backslash is a legal byte in a Linux filename. `internal/metadata` builds on-disk
`.meta` names straight from the virtual path, so such a name is persisted as-is, and
opening that file through the FUSE layer deadlocks: the kernel hangs in
`invalidate_inode_pages2` (`folio_wait_bit_common`) and the calling process goes into
uninterruptible D-state. A host reboot is the only recovery.

The two layers disagree. `nzbfilesystem.normalizePath` already folds backslashes to
forward slashes, but the metadata service does not, so writes land under one key and
lookups arrive under another. Still reproducible on `b6a4a977`: `WriteFileMetadata`
calls `filepath.Base(virtualPath)` on the raw path.

#849 does not cover this. It fixes separators at the RPC boundary and says outright that
a backslash is a legal POSIX filename character left alone by design. This is about a
byte that must never reach an on-disk name in the first place.

### What is here

@javi11's `fa243c60` from #685 is unchanged in the history: the normalization lives at a
single chokepoint, `metaFilePath` / `metaDirPath`, rather than being repeated at each
method top, so a future caller cannot forget it. Method-top normalization is kept only
where the local variable is also a `liteCache` key or a prefix-scan argument.

Since #685 stalled, #875 replaced `DeleteFileMetadataWithSourceNzb` with
`DeleteFileMetadata(ctx, virtualPath)`. The chokepoint is ported onto that, and the
regression test that called the old signature is rewritten.

One extra fix, from an external review of this branch. The `DeleteDirectory` cache purge
built its prefix as `virtualPath + filepath.Separator`, which misses in two ways once
paths are normalized. Cache keys always use forward slashes, so on Windows the prefix
reads `movies\` against keys like `movies/file.mkv` and matches nothing at all. And a
trailing separator doubles up, turning `movies/` into `movies//`, which also matches
nothing. Either way the directory goes from disk while its entries stay cached, and the
next read is served metadata for a file that is gone.

### Tests

`assertNoBackslashPersisted` walks the metadata root and fails on any persisted name
containing a backslash. Regression tests cover the write, rename and delete paths plus
the directory cache purge, and each was checked against the unpatched code to confirm it
actually fails there.

`go build ./...` and `go test ./...` pass on Linux with Go 1.26. The one failure I saw,
`TestIsWritable` in `internal/updater`, reproduces on unmodified `main` in the same
container and is an artifact of running the suite as root, where mode `0o400` is still
writable.

### Behaviour changes beyond backslash normalization

Two, both surfaced in review and worth being explicit about.

**Reads truncate now.** `ReadFileMetadata`, `ReadFileMetadataLite`,
`readFileMetadataLiteFull`, `DeleteFileMetadata`, `RenameFileMetadata` and
`GetMetadataFilePath` previously built their path from the raw `filepath.Base`; only
`WriteFileMetadata` and `FileExists` truncated. Routing everything through `metaFilePath`
means they all truncate. Mostly that is a fix: the stat-then-delete-then-write pattern in
`multifile/processor.go`, `rar/aggregator.go` and `sevenzip/aggregator.go` used to stat an
un-truncated path that could not exist, so a re-import of a long name skipped its
pre-delete and leaked the store refcount. The hazard it introduces is that two names
sharing their first bytes and extension now read as the same file where the read previously
missed; `TestLongFilename_SharedPrefixCollides` documents that, and closing it needs an
on-disk layout change.

**Long names can be written at all now.** `truncateFilename` budgeted 250 bytes for the
base "to leave room for .meta" without subtracting the file's own extension, and the
atomic-write temp file was built from the already-maximal truncated name. Anything over
roughly 235 bytes failed with "file name too long". Both are fixed here because the test
asked for above could not otherwise be written. Only names that could not be written are
affected, so nothing on disk needs migrating.

### Migration

`MigrateBackslashPaths` relocates metadata already written under an un-normalized name,
which the normalization would otherwise strand: invisible to `Readdir`, dropped by library
sync without being flagged corrupted, undeletable, skipped by the v2→v3 scan, and leaking
its store refcount. It runs once from `initializeMetadata`, is idempotent, and never
overwrites a live file.

### DIFF
--- a/internal/metadata/service.go
+++ b/internal/metadata/service.go
@@ -8,6 +8,7 @@ import (
 	"io/fs"
 	"log/slog"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 	"time"
@@ -133,34 +134,65 @@ func (ms *MetadataService) readStoreRef(metaFilePath string) string {
 	return fm.StoreRef
 }
 
-// truncateFilename truncates the filename if it's too long to prevent filesystem issues
-// when creating .meta files. Keeps filename under 250 characters.
+// normalizeVirtualPath canonicalizes a virtual path so every caller ends up
+// with exactly one on-disk name and one liteCache key for the same file.
+// A backslash is a legal Linux filename byte but deadlocks the FUSE
+// page-cache layer on open() (issue #660), so it's folded to forward slash;
+// path.Clean (not filepath.Clean, which would turn it back into a backslash
+// on Windows) then collapses "//" and "..", and the leading slash is
+// stripped so "/x" and "x" collapse to one key.
+func normalizeVirtualPath(virtualPath string) string {
+	v := strings.ReplaceAll(virtualPath, "\\", "/")
+	v = path.Clean(v)
+	v = strings.TrimPrefix(v, "/")
+	if v == "." {
+		v = ""
+	}
+	return v
+}
+
+// truncateFilename truncates the base so base+ext+".meta" fits within the
+// 255-byte name-component limit most filesystems enforce.
 func (ms *MetadataService) truncateFilename(filename string) string {
+	const maxNameLen = 255
+	const metaSuffixLen = len(".meta")
+
 	fileExt := filepath.Ext(filename)
-	filename = strings.TrimSuffix(filename, fileExt)
+	base := strings.TrimSuffix(filename, fileExt)
 
-	const maxLen = 250 // Leave room for .meta extension
+	maxBaseLen := max(maxNameLen-metaSuffixLen-len(fileExt), 0)
 
-	if len(filename) <= maxLen {
-		return filename + fileExt
+	if len(base) <= maxBaseLen {
+		return base + fileExt
 	}
 
-	// Simply truncate to maxLen
-	return filename[:maxLen] + fileExt
+	return base[:maxBaseLen] + fileExt
+}
+
+// metaFilePath is the single chokepoint for on-disk .meta file path
+// construction: every caller gets the same normalized, truncated path.
+func (ms *MetadataService) metaFilePath(virtualPath string) string {
+	v := normalizeVirtualPath(virtualPath)
+	dir := filepath.Join(ms.rootPath, filepath.Dir(v))
+	name := ms.truncateFilename(filepath.Base(v))
+	return filepath.Join(dir, name+".meta")
+}
+
+// metaDirPath is the single chokepoint for on-disk directory path
+// construction.
+func (ms *MetadataService) metaDirPath(virtualPath string) string {
+	return filepath.Join(ms.rootPath, normalizeVirtualPath(virtualPath))
 }
 
 // WriteFileMetadata writes file metadata to disk
 func (ms *MetadataService) WriteFileMetadata(virtualPath string, metadata *metapb.FileMetadata) error {
-	// Ensure the directory exists
-	metadataDir := filepath.Join(ms.rootPath, filepath.Dir(virtualPath))
+	virtualPath = normalizeVirtualPath(virtualPath)
+
+	metadataPath := ms.metaFilePath(virtualPath)
+	metadataDir := filepath.Dir(metadataPath)
 	if err := os.MkdirAll(metadataDir, 0755); err != nil {
 		return fmt.Errorf("failed to create metadata directory: %w", err)
 	}
-
-	// Create metadata file path (filename + .meta extension)
-	filename := filepath.Base(virtualPath)
-	truncatedFilename := ms.truncateFilename(filename)
-	metadataPath := filepath.Join(metadataDir, truncatedFilename+".meta")
 
 	// Sidecar ID handling for compatibility
 	// We don't write NzbdavId to the proto to maintain compatibility with versions that don't have field 14.
@@ -200,9 +232,10 @@ func (ms *MetadataService) WriteFileMetadata(virtualPath string, metadata *metap
 		writeData = raw
 	}
 
-	// Write atomically using a uniquely-named temporary file so concurrent
-	// writes to the same final path don't race on the same .tmp name.
-	tmpFile, err := os.CreateTemp(metadataDir, "."+truncatedFilename+".*.tmp")
+	// Write atomically via a uniquely-named temp file so concurrent writes
+	// don't race on the same name. Not built from the truncated filename:
+	// that would risk exceeding the name-length limit truncation exists for.
+	tmpFile, err := os.CreateTemp(metadataDir, ".meta-*.tmp")
 	if err != nil {
 		metadata.NzbdavId = nzbdavId
 		return fmt.Errorf("failed to create temporary metadata file: %w", err)
@@ -332,10 +365,8 @@ func (ms *MetadataService) WriteFileMetadataAuto(ctx context.Context, virtualPat
 // caller's handle. As a side effect, the lightweight projection is cached so
 // subsequent Readdir/Stat calls are fast without a disk read.
 func (ms *MetadataService) ReadFileMetadata(virtualPath string) (*metapb.FileMetadata, error) {
-	// Create metadata file path
-	filename := filepath.Base(virtualPath)
-	metadataDir := filepath.Join(ms.rootPath, filepath.Dir(virtualPath))
-	metadataPath := filepath.Join(metadataDir, filename+".meta")
+	virtualPath = normalizeVirtualPath(virtualPath)
+	metadataPath := ms.metaFilePath(virtualPath)
 
 	// Read file
 	data, err := os.ReadFile(metadataPath)
@@ -421,15 +452,15 @@ const liteScanBytes = 4096
 // NestedSources/SegmentData slices. Falls back to a full read in the rare
 // case the partial buffer doesn't cover the lite fields.
 func (ms *MetadataService) ReadFileMetadataLite(virtualPath string) (*FileMetadataLite, error) {
+	virtualPath = normalizeVirtualPath(virtualPath)
+
 	// Check lite cache first
 	if cached, ok := ms.liteCache.Get(virtualPath); ok {
 		return cached, nil
 	}
 
 	// Cache miss — read the head of the file and scan wire-format fields.
-	filename := filepath.Base(virtualPath)
-	metadataDir := filepath.Join(ms.rootPath, filepath.Dir(virtualPath))
-	metadataPath := filepath.Join(metadataDir, filename+".meta")
+	metadataPath := ms.metaFilePath(virtualPath)
 
 	f, err := os.Open(metadataPath)
 	if err != nil {
@@ -531,9 +562,8 @@ func parseLiteFields(buf []byte) (*FileMetadataLite, bool) {
 // partial-read scan in ReadFileMetadataLite fails to locate the lite
 // fields within liteScanBytes.
 func (ms *MetadataService) readFileMetadataLiteFull(virtualPath string) (*FileMetadataLite, error) {
-	filename := filepath.Base(virtualPath)
-	metadataDir := filepath.Join(ms.rootPath, filepath.Dir(virtualPath))
-	metadataPath := filepath.Join(metadataDir, filename+".meta")
+	virtualPath = normalizeVirtualPath(virtualPath)
+	metadataPath := ms.metaFilePath(virtualPath)
 
 	data, err := os.ReadFile(metadataPath)
 	if err != nil {
@@ -563,25 +593,19 @@ func (ms *MetadataService) readFileMetadataLiteFull(virtualPath string) (*FileMe
 
 // FileExists checks if a metadata file exists for the given virtual path
 func (ms *MetadataService) FileExists(virtualPath string) bool {
-	filename := filepath.Base(virtualPath)
-	truncatedFilename := ms.truncateFilename(filename)
-	metadataDir := filepath.Join(ms.rootPath, filepath.Dir(virtualPath))
-	metadataPath := filepath.Join(metadataDir, truncatedFilename+".meta")
-
-	_, err := os.Stat(metadataPath)
+	_, err := os.Stat(ms.metaFilePath(virtualPath))
 	return err == nil
 }
 
 // DirectoryExists checks if a metadata directory exists
 func (ms *MetadataService) DirectoryExists(virtualPath string) bool {
-	metadataDir := filepath.Join(ms.rootPath, virtualPath)
-	info, err := os.Stat(metadataDir)
+	info, err := os.Stat(ms.metaDirPath(virtualPath))
 	return err == nil && info.IsDir()
 }
 
 // ListDirectory lists all metadata files in a directory
 func (ms *MetadataService) ListDirectory(virtualPath string) ([]string, error) {
-	metadataDir := filepath.Join(ms.rootPath, virtualPath)
+	metadataDir := ms.metaDirPath(virtualPath)
 
 	entries, err := os.ReadDir(metadataDir)
 	if err != nil {
@@ -607,7 +631,7 @@ func (ms *MetadataService) ListDirectory(virtualPath string) ([]string, error) {
 // file names from a single os.ReadDir call. This is used by Readdir to avoid
 // two separate directory reads.
 func (ms *MetadataService) ListDirectoryAll(virtualPath string) (dirs []fs.FileInfo, fileNames []string, err error) {
-	metadataDir := filepath.Join(ms.rootPath, virtualPath)
+	metadataDir := ms.metaDirPath(virtualPath)
 
 	entries, err := os.ReadDir(metadataDir)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


javi11 (commented):
Reviewed for regression risk. **The change is sound, and metadata really was the odd layer out** — every other layer already folds backslashes: `nzbfilesystem.normalizePath` (`internal/nzbfilesystem/types.go:64`), `database.normalizeHealthPath` (`internal/database/health_repository.go:40-44`, funnelled through by all `file_path` writers/matchers), and the importer at `internal/importer/filesystem/utils.go:151` plus both archive aggregators (`rar/aggregator.go:270`, `sevenzip/aggregator.go:211`). So `foo\bar.mkv` → directory `foo` + file `bar.mkv` is what the importer has always produced; this makes metadata agree rather than diverge.

Also checked and clean: no DB/metadata divergence at any of the DB→metadata seams (`arrs_handlers.go:459`, `health/worker.go:473,543,709`, `health_handlers.go:233,317`); the synthesized directory is not a phantom (`Dir.Lookup` → `nzbfs.Stat` → `normalizePath` → `DirectoryExists` resolves it, and the health row matches); `.tmp` and every `.id` site derive from the same truncated+normalized base, and nothing in the repo *writes* `.id`, so no sidecar leak; all six `GetMetadataFilePath`/`GetMetadataDirectoryPath` callers are safe — the only fs→virtual round trip (`api/file_handlers.go:445` → `:496`) walks from `metaDirPath("/")` so the `Rel` base matches, and `truncateFilename` is idempotent; and the `DeleteDirectory` rewrite is strictly better — the new `cacheDir != ""` guard skips the purge exactly for the `""`/`"/"`/all-slash inputs that the HARD SAFETY check at `service.go:874-877` rejects before any `RemoveAll`, so there is no input where the tree is deleted but the cache is left populated.

Inline comments cover four points. Two more can't be anchored because they're outside the diff:

### `DirectoryModTime` still bypasses the chokepoint

`internal/metadata/service.go:661` is the one remaining raw join inside the package:

```go
info, err := os.Stat(filepath.Join(ms.rootPath, virtualPath))
```

Latent rather than live — its callers at `metadata_remote_file.go:521,753` have already been through `normalizePath` — but it makes the "single chokepoint" claim in the new `metaFilePath` doc comment untrue. Should be `ms.metaDirPath(virtualPath)`.

### Two `.meta` constructors outside the package bypass it entirely

Both hand-build the path with **neither normalization nor truncation**:

- `internal/importer/postprocessor/symlink_creator.go:59,64` — `metadataPath := filepath.Join(cfg.Metadata.RootPath, strings.TrimPrefix(originalResultingPath, "/"))` then `metaFile := metadataPath + ".meta"`
- `internal/importer/postprocessor/strm_generator.go:50,55` — identical pattern

So symlink and `.strm` creation miss for any path the metadata layer normalized, and also for names over 250 bytes (which the writer truncates but these do not). Both should call `metadataService.GetMetadataFilePath`. This is the same class of drift the chokepoint was introduced to prevent, just one package over.

### Test gaps

The four new tests are genuine red tests on `main` (verified). Not covered: a `>250`-char name anywhere in the suite; a pre-existing backslash `.meta` already on disk; leading-slash cache keys (all four tests use relative paths, production passes `/`-prefixed); `ListDirectory` once a backslash name has become a nested directory; `DeleteDirectory("")` / `("/")`.

internal/metadata/service.go
```diff
@@ -144,18 +144,49 @@ func (ms *MetadataService) truncateFilename(filename string) string {
...
+// path. It is the single chokepoint for file-path construction: backslash
+// normalization (issue #660) and filename truncation both happen here, so no
+// caller — current or future — can persist or serve a backslash or an over-long
+// name.
+func (ms *MetadataService) metaFilePath(virtualPath string) string {
+	virtualPath = normalizeVirtualPath(virtualPath)
+	dir := filepath.Join(ms.rootPath, filepath.Dir(virtualPath))
+	name := ms.truncateFilename(filepath.Base(virtualPath))
+	return filepath.Join(dir, name+".meta")
+}
```
javi11, line 179: **Pre-existing backslash `.meta` files are silently orphaned by this, and nothing migrates them.**

There is no `.meta` renaming routine anywhere in the repo. The only metadata migration is the v1→v3 store conversion (`internal/metadata/migration.go:82`), which is manually triggered from the panel (`internal/api/server.go:311-314`), not run at startup. So on an install that already hit #660 the on-disk file is `…/foo\bar.mkv.meta`, every accessor now resolves to `…/foo/bar.mkv.meta`, and the legacy file becomes:

- **Invisible.** `Readdir` reads raw on-disk names via `ListDirectoryAll`, joins them (`internal/nzbfilesystem/metadata_remote_file.go:702`) and calls `ReadFileMetadataLite`, which now normalizes and misses → `continue` at `:703-705`. The file just disappears from the library.
- **Dropped by library sync without even being flagged corrupted.** `metaPathToMountRelativePath` (`internal/health/library_sync.go:1128`) uses `filepath.ToSlash`, which is a **no-op on Linux**, so the mount-relative key keeps the literal backslash. It reaches `ReadFileMetadata`, which returns `(nil, nil)` for a missing file — so the `RegisterCorruptedFile` branch at `:751-762` is never taken. No log, no health row.
- **Undeletable and refcount-leaking.** `DeleteFileMetadata` targets the normalized path, so neither the `.meta` nor its `.id` is ever removed, and `readFileMetadataStoreRef` (`service.go:791`) reads the new path, gets `""`, so the v3 store's refcount is never decremented.
- **Excluded from the v2→v3 migration without being reported.** `ScanLegacyMetas` derives `virtualPath` from `filepath.Rel` with the backslash intact (`migration_scan.go:54-58`), then calls `ms.ReadFileMetadata` at `:61`; that now misses and the walk does a bare `return nil` — dropped from the scan, not counted as a failure.

Not a deadlock any more, which is the point of the PR, but "files vanish from the library with no log line" is a weak replacement for it, and the affected population is exactly the #660 reporters.

Suggested: a one-shot sweep that walks the metadata root and `os.Rename`s any `.meta` (and its `.id`) whose name contains `\` into its normalized nested location. Minimum acceptable would be a `WARN` naming each such file so a user can act on it.

internal/metadata/service.go
```diff
@@ -327,10 +358,8 @@ func (ms *MetadataService) WriteFileMetadataAuto(ctx context.Context, virtualPat
 // caller's handle. As a side effect, the lightweight projection is cached so
 // subsequent Readdir/Stat calls are fast without a disk read.
 func (ms *MetadataService) ReadFileMetadata(virtualPath string) (*metapb.FileMetadata, error) {
-	// Create metadata file path
-	filename := filepath.Base(virtualPath)
-	metadataDir := filepath.Join(ms.rootPath, filepath.Dir(virtualPath))
-	metadataPath := filepath.Join(metadataDir, filename+".meta")
+	virtualPath = normalizeVirtualPath(virtualPath)
+	metadataPath := ms.metaFilePath(virtualPath)
```
javi11, line 369: **Truncation is now applied on the read side too — a real behavior change beyond backslash normalization, and it isn't mentioned in the PR description or covered by a test.**

`ReadFileMetadata`, `ReadFileMetadataLite`, `readFileMetadataLiteFull`, `DeleteFileMetadata`, `RenameFileMetadata` and `GetMetadataFilePath` all previously built the path from the **raw** `filepath.Base`; only `WriteFileMetadata` and `FileExists` truncated. Routing them through `metaFilePath` means every one of them truncates now.

Mostly this is a **fix**, and a good one: the stat-then-delete-then-write pattern at `multifile/processor.go:142`, `rar/aggregator.go:397` and `sevenzip/aggregator.go:332` previously stat'd an un-truncated path that never existed, so re-imports of >250-byte names skipped the pre-delete entirely and never decremented the store refcount. That now works.

The new hazard is worth a deliberate call: `truncateFilename` is a plain 250-**byte** cut, so two names sharing the first 250 bytes and the same extension now *read* as the same file, where previously the read simply missed. That turns an ENOENT into silently-wrong metadata. (It also cuts mid-rune on UTF-8, though that predates this PR on the write side.)

There is currently no test anywhere in the suite with a >250-char name — worth adding one now that reads depend on the same truncation the writer applies.

internal/metadata/service.go
```diff
@@ -144,18 +144,49 @@ func (ms *MetadataService) truncateFilename(filename string) string {
...
+// A backslash (0x5C) is a legal byte in a Linux filename, but a file whose name
+// contains one deadlocks the FUSE page-cache layer on open() — the kernel hangs
+// indefinitely in invalidate_inode_pages2 (folio_wait_bit_common) and the calling
+// process goes uninterruptible D-state, recoverable only by rebooting the host
+// (issue #660). Forward slash matches the convention already used by
+// nzbfilesystem.normalizePath so metadata writes and database lookups resolve to
+// the same key.
+func normalizeVirtualPath(virtualPath string) string {
+	return strings.ReplaceAll(virtualPath, "\\", "/")
+}
```
javi11, line 170: Minor, and pre-existing rather than introduced here — but it's the same class of bug as the `DeleteDirectory` purge fix in this PR, so worth noting while you're in here.

Unlike `nzbfilesystem.normalizePath` (`internal/nzbfilesystem/types.go:52-78`), this neither `filepath.Clean`s nor canonicalizes a leading `/`. `liteCache` is therefore still keyed on whatever shape the caller passed: `"/movies/x.mkv"` and `"movies/x.mkv"` remain two distinct entries for the same file.

The live consequence is at `migration_convert.go:116`, which removes under a no-leading-slash key (`lm.VirtualPath` comes from `filepath.Rel` at `migration_scan.go:58`) while every FUSE reader caches with the leading slash — so that eviction can never hit the entry the serving path is actually using.

Not a blocker; folding `Clean` + leading-slash canonicalization into `normalizeVirtualPath` would close it, if you want the chokepoint to fully own path shape.

internal/metadata/service.go
```diff
@@ -1000,9 +1036,11 @@ func (ms *MetadataService) cleanupEmptyDirsRecursive(path string, protected []st
 
 // MoveToCorrupted moves a metadata file to a special corrupted directory for safety
 func (ms *MetadataService) MoveToCorrupted(ctx context.Context, virtualPath string) error {
+	virtualPath = normalizeVirtualPath(virtualPath)
 	ms.liteCache.Remove(virtualPath)
 
-	// Normalize path and remove leading slashes to ensure it joins correctly
+	// Remove leading slashes to ensure it joins correctly. virtualPath is already
```
javi11: `MoveToCorrupted` normalizes at the top, then re-implements the chokepoint rather than calling it — `TrimPrefix` / `FromSlash` / `Dir` / `Base` / `truncateFilename` / `Join` at `:1042-1049`, plus the same construction again for `targetPath` at `:1065`.

The result is *equivalent* to `metaFilePath` today (`FromSlash` and `TrimPrefix` are no-ops for correctness since `Join` cleans anyway), so there's no bug. But it is a second copy of the rule that a future change to `metaFilePath` will not reach, which is precisely what the new chokepoint exists to prevent. Worth collapsing to `ms.metaFilePath(virtualPath)` and deriving `dir` from `filepath.Dir` of the result.

PR #879, opened 2026-08-31
fix(rclone): obscure the webdav password when creating the internal mount config

### LINKED ISSUE #691: Internal rclone mount 401s when login_required is enabled (generated rclone config stores the webdav password un-obscured)

firestaerter3:
Fixes #691.

`createConfig` calls `config/create` without the `obscure` opt, so rclone decides for
itself whether the password it was handed is already obscured: it tries to reveal the
value, and stores it verbatim when that succeeds. For a plaintext password that happens
to be revealable, the stored value then reveals to something else, and the internal
mount authenticates against AltMount's own WebDAV with the wrong credential. That is a
401 when `auth.login_required` is on.

@javi11, this is why it worked when you checked on #691. A password with no particular
shape round-trips fine either way. A base64-shaped one, which is what most generated
passwords look like, does not.

Reproduced on rclone v1.75.0. For each row: `config/create` with the given password, then
`rclone reveal` on what ended up in `rclone.conf`.

| password handed to `config/create` | `opt.obscure` | reveals back to the original |
| --- | --- | --- |
| ordinary string | absent | yes |
| ordinary string | `true` | yes |
| itself a valid obscured string | absent | **no** |
| itself a valid obscured string | `true` | yes |

Row three is the bug. The password is stored un-obscured, so every later read de-obscures
it into a different string.

The password still goes over as plaintext; `obscure` only tells rclone to do the obscuring
rather than guess. Recreating the remote therefore supplies plaintext again and repairs a
previously mis-stored value instead of double-encoding it.

@WhispersOfJ independently reproduced this on #691 and had a fix in #792, which was closed
when it got folded into #799. This is that change on its own, plus a test asserting the RC
body carries `opt.obscure` and still sends the password in plaintext. The test fails
against the current `createConfig`.

### DIFF
--- a/pkg/rclonecli/client.go
+++ b/pkg/rclonecli/client.go
@@ -496,6 +496,13 @@ func (m *Manager) createConfig(ctx context.Context, configName, webdavURL string
 				"user":            user,
 				"pass":            pass,
 			},
+			// Obscure the plaintext webdav password before rclone stores it.
+			// Without this rclone keeps it as-is, then fails to reveal it and
+			// sends a garbled password to the local webdav → 401 when
+			// auth.login_required is true (the whole native-mount blocker).
+			"opt": map[string]any{
+				"obscure": true,
+			},
 		},
 	}
 
--- a/pkg/rclonecli/client.go
+++ b/pkg/rclonecli/client.go
@@ -496,17 +496,9 @@ func (m *Manager) createConfig(ctx context.Context, configName, webdavURL string
 				"user":            user,
 				"pass":            pass,
 			},
-			// Tell rclone the password above is plaintext.
-			//
-			// Without this, config/create guesses: it tries to reveal the value
-			// and, if that succeeds, assumes it was already obscured and stores
-			// it verbatim. The guess is wrong for any plaintext password that
-			// happens to be revealable — base64-shaped strings, which is what a
-			// randomly generated password usually looks like. Such a password is
-			// then stored un-obscured, reveals to a different string on every
-			// later read, and the mount authenticates against altmount's own
-			// WebDAV with the wrong credential: 401 when auth.login_required is
-			// on. Passing obscure explicitly removes the guess.
+			// obscure tells rclone the password is plaintext. Without it,
+			// config/create guesses, and guesses wrong for any password that is
+			// itself revealable, storing it verbatim (#691).
 			"opt": map[string]any{
 				"obscure": true,
 			},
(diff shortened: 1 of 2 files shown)


pkg/rclonecli/client.go
```diff
@@ -496,6 +496,20 @@ func (m *Manager) createConfig(ctx context.Context, configName, webdavURL string
 				"user":            user,
 				"pass":            pass,
 			},
+			// Tell rclone the password above is plaintext.
```
javi11: Can you remove that huge comment blocks?

firestaerter3: Trimmed both to three lines. The detail is in #691 and the PR description if anyone needs it.

javi11 (approved):

Issue #886, opened 2026-09-01
bug: failed archive import rollback runs os.RemoveAll on release folder, deleting concurrent/sibling imports and deadlocking FUSE (ffprobe in D-state)


actuallyevan:
**NOTE:**
Obviously, AI helped me write the sample PR and the issue, but I actually ran into this problem with my own stack.

In my opinion, the issue below describes the best way to fix the problem, but another thing to note is that Altmount seems to trigger **duplicate searches** when repair is triggered on files that have fully imported into Sonarr's library. **This can be a catalyst for the problem, although it might not be worth fixing.**

Here is a draft PR that could solve the problem (although there could be issues with it): https://github.com/javi11/altmount/pull/887

### Summary

When an archive (RAR or 7z) import fails validation or processing, its failure rollback in `cleanupWrittenPaths` deletes the entire release directory using `os.RemoveAll`.

Because import directories are deterministic (`/category/ReleaseName`), if a duplicate or sibling download of the same release is running concurrently (or just completed), the failing job completely wipes out the healthy files. If an ARR (Sonarr/Radarr) has already begun importing the healthy download via the FUSE/rclone mount, deleting the backing metadata file mid-read stalls the mount and puts `ffprobe` into an unkillable Linux kernel **`D` state (`TASK_UNINTERRUPTIBLE`)**, permanently freezing the ARR import worker.

### Root Cause

There is an architectural inconsistency between how `multifile` and `archive` processors track written paths:

* **In `multifile/processor.go`**: As files are written, their exact virtual paths are tracked in `writtenPaths`. On failure, `cleanupWrittenPaths` deletes **only the files that specific job wrote**. It never deletes directories with `os.RemoveAll`. 
* **In `archive/rar/processor.go:947` (and 7z:1080)**: Before any extraction or validation occurs, it registers: `writtenPaths := []string{"DIR:" + nzbFolder}`

And on failure, cleanupWrittenPaths blindly runs os.RemoveAll on that directory via s.metadataService.DeleteDirectory(dirPath).

### Expected Behavior
A failed archive download rollback should:

1. Only delete the specific files that that job actually wrote.
2. Only remove the release directory if it is completely empty (os.Remove, never os.RemoveAll).
3. Never touch, delete, or disrupt files belonging to concurrent or pre-existing imports in the same directory.

PR #887, opened 2026-09-01
bug: failed archive import rollback runs os.RemoveAll on release folder, deleting concurrent/sibling imports and deadlocking FUSE (ffprobe in D-state)

actuallyevan:
Draft PR for https://github.com/javi11/altmount/issues/886

Not sure if this will break other stuff.


internal/importer/processor.go
```diff
@@ -979,7 +980,13 @@ func (proc *Processor) processRarArchive(
 
 		releaseDate := archiveFiles[0].ReleaseDate.Unix()
 
+		var writtenMu sync.Mutex
 		err := rar.ProcessArchive(ctx, rar.ProcessArchiveOptions{
+			OnMetadataWritten: func(vp string) {
```
javi11, line 1186: **HIGH: `ProcessRegularFiles` writes are untracked → this job's own files leak on rollback.**

Both archive paths discard the written-paths return value (`if _, err := multifile.ProcessRegularFiles(...)` at ~L957 and ~L1095). Under the old semantics that was fine — the `DIR:` RemoveAll swept everything. Under the new "delete dir only if empty" semantics, when the subsequent archive processing fails, those regular-file `.meta` files (nfo/srt/sample siblings) keep the dir non-empty, so the failed import's own metadata is never cleaned and orphaned entries appear in the VFS.

Fix: capture the return and append to `writtenPaths` — the non-archive path at ~L860 already does this. This is what breaks `TestCleanupWrittenPaths_DeletesDirectory` / `TestCleanupWrittenPaths_MixedPaths`.

internal/importer/service.go
```diff
@@ -1419,6 +1410,21 @@ func (s *Service) cleanupWrittenPaths(ctx context.Context, itemID int64, paths [
...
+		if delErr := s.metadataService.DeleteDirectoryIfEmpty(dirPath); delErr != nil {
+			s.log.DebugContext(ctx, "Preserved metadata directory after import failure (contains files from another import)",
+				"queue_id", itemID,
+				"dir", dirPath)
+		} else {
+			s.log.DebugContext(ctx, "Cleaned up empty metadata directory after import failure",
+				"queue_id", itemID,
+				"dir", dirPath)
+		}
+		s.cleanupHealthRecords(ctx, itemID, dirPath)
```
javi11, line 1493: **HIGH: `cleanupHealthRecords` on a *preserved* directory deletes the sibling import's health records.**

This runs unconditionally, and it deletes unvalidated health records **by path prefix** for the whole directory. When the dir is preserved precisely because a concurrent/sibling import owns files in it, this wipes the sibling's pending health records — reintroducing the cross-import interference this PR exists to fix, just in the health table instead of the filesystem.

Fix: call `cleanupHealthRecords` for the dir only in the successful-delete branch; the failed job's per-file records are already cleaned in pass 1.

internal/importer/service.go
```diff
@@ -1419,6 +1410,21 @@ func (s *Service) cleanupWrittenPaths(ctx context.Context, itemID int64, paths [
 			s.cleanupHealthRecords(ctx, itemID, p)
 		}
 	}
+
+	// Pass 2: Now that this job's files have been removed, delete the directory
+	// ONLY if it is empty. If sibling files exist, it is preserved unharmed.
+	for _, dirPath := range dirPaths {
+		if delErr := s.metadataService.DeleteDirectoryIfEmpty(dirPath); delErr != nil {
```
javi11: **MEDIUM: error conflation — real errors reported as "contains files from another import".**

Any error from `DeleteDirectoryIfEmpty` (EACCES, EIO, the root-safety block) lands here and is logged at Debug with a message asserting sibling files exist. Distinguish `errors.Is(err, syscall.ENOTEMPTY)` (Debug, expected) from other errors (Warn). Note Windows returns `ERROR_DIR_NOT_EMPTY`, so if cross-platform matters, match both or probe emptiness explicitly.

internal/metadata/service.go
```diff
@@ -886,6 +886,33 @@ func (ms *MetadataService) DeleteDirectory(virtualPath string) error {
 	return nil
 }
 
+// DeleteDirectoryIfEmpty deletes a metadata directory only if it contains no files or subdirectories.
+// Returns an error (ENOTEMPTY) if the directory is not empty, guaranteeing concurrent/sibling files are never deleted.
+func (ms *MetadataService) DeleteDirectoryIfEmpty(virtualPath string) error {
```
javi11, line 888: **MEDIUM: duplicated safety-block + cache-purge logic.**

This copy-pastes the root-safety check and the liteCache prefix-purge loop from `DeleteDirectory`. Extract shared helpers (e.g. `ms.assertNotRoot(path)`, `ms.purgeCachePrefix(virtualPath)`).

LOW (informational, pre-existing in `DeleteDirectory` too): no guard against `..` in `virtualPath` escaping `rootPath` — paths are internally generated, so just noting it.

javi11 (changes requested):
The approach is right — replacing `os.RemoveAll` on the shared release folder with an atomic `os.Remove` (fails ENOTEMPTY) is exactly the correct primitive, and the write-tracking callback covers the aggregators' single write site. But two HIGH issues need fixing before merge, and 2 existing tests fail (`TestCleanupWrittenPaths_DeletesDirectory`, `TestCleanupWrittenPaths_MixedPaths`) — they encode the old contract and need updating to the new one, plus new coverage for the sibling-preservation path. Details inline.

✅ Security-wise this strictly reduces the destructive-deletion blast radius, and the atomic `os.Remove` avoids the check-then-delete race a ReadDir-then-RemoveAll approach would have had.

